PR #1, opened 2014-08-18
CYB-15: Design and Implementaion of Java API for ingesting into stream

ashvayka:
Implementation and main tests.


nitinmotgi:
Awesome work Andrew. Don't be alarmed by code review comments to follow - it's normal.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,176 @@
...
+   * @throws IOException in case if entity content is not available
+   */
+  public static JsonObject getEntityAsJsonObject(HttpEntity httpEntity) throws IOException {
+    JsonObject result;
+    if (httpEntity != null && httpEntity.getContent() != null) {
+      InputStream entityInputStream = httpEntity.getContent();
+      String content = null;
+      Reader reader = null;
+      try {
+        reader = new InputStreamReader(entityInputStream);
```
chtyim, line 138: You can move this line out of the try block. If you do so, you don't need to check for `null` in finally, making the code cleaner.

``` java
Reader reader = new InputStreamReader(httpEntity.getContent(), charset);
try {
  content = ...
} finally {
  reader.close();
}
```

chtyim, line 138: Also, when creating `Reader` from `InputStream`, always supply a charset, otherwise it'll be using the platform charset. The charset should be extracted from the response content-type if available, or UTF-8 as this is assuming the content is in JSON.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,176 @@
...
+
+  private final RestClientConnectionConfig config;
+  private final String baseUrl;
+  private final CloseableHttpClient httpClient;
+
+  public RestClient(RestClientConnectionConfig config, HttpClientConnectionManager connectionManager) {
+    this.config = config;
+    this.baseUrl = config.isSsl() ? HTTPS_PROTOCOL : HTTP_PROTOCOL + PROTOCOL_POSTFIX + config.getHost() +
+      COLON + config.getPort() + FORWARD_SLASH + config.getVersion() + FORWARD_SLASH;
+    this.httpClient = HttpClients.custom().setConnectionManager(connectionManager).build();
```
chtyim, line 71: If possible, I would prefer not to use `httpclient` library to reduce dependency. The standard Java `URL`, `URLConnection` and `HttpConnectionURL` class should be able to provides all functionality needed (including persisted connection), with about the same amount of code. Is there any specific functionality that you needed not found in the standard java class?

checkstyle.xml
```diff
@@ -0,0 +1,392 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
abaranau, line 1: this file is copied from cdap repo, right?

pom.xml
```diff
@@ -0,0 +1,207 @@
...
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
```
abaranau, line 64: let's take it out until we start using it. We still figuring out which lib to use as standard way for mocking in unit-tests

abaranau, line 64: I don't think you use it yet, right?

checkstyle.xml
ashvayka, line 1: Yes, correct

pom.xml
```diff
@@ -0,0 +1,207 @@
...
+    <gson.version>2.2.4</gson.version>
+    <rs-api.version>2.0</rs-api.version>
+    <jersey.version>2.4.1</jersey.version>
+    <slf4j.version>1.7.6</slf4j.version>
+    <mockito.version>1.9.5</mockito.version>
+    <junit.version>4.11</junit.version>
+  </properties>
+
+
+  <dependencies>
```
abaranau, line 47: define dependencyManagement in parent pom, and actual dependencies in poms of modules that need them.

stream-client/stream-client-java/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+
+  <artifactId>stream-client-java</artifactId>
+  <packaging>jar</packaging>
+  <name>Stream Client Java API</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>${http.component.version}</version>
```
abaranau, line 37: define version to be used in parent pom in dependency management. The idea is that parent pom has all dependencies versions defined, while not declaring any actual dependencies. Then module's pom.xml defines actual dependencies without explicit versions. Thus version defined in parent pom is used everywhere in child modules.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
ashvayka, line 71: The reason about HttpClient is mostly about performance (however it is also much simpler to use). We use http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html to optimize performance and allow parallel execution of requests per one host. This method: http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#setDefaultMaxPerRoute%28int%29 allow to set max parallel connections per one host. This will help to upload multiple events (batches in future) per one host simultaneously.

pom.xml
ashvayka, line 64: We already use in File Tailer app that we are going to push tomorrow morning. There is a lot of functionality covered in those test. Is there a chance to review that code and make this decision afterwards?

awholegunch:
A package-info.java with documentation in it would be helpful and is usually included in all our packages. Please note our spelling for common Cask elements: Stream, Flow, MapReduce Job, Workflow, etc; these all start with an initial capital letter.

sreevatsanraman:
Can this be closed? This is a stale one and PR 2 is the one that will replace this.

PR #2, opened 2014-08-19
CYB #15: Design and Implementaion of Java API for ingesting into stream

ashvayka:
Implementation and main tests.

This is a duplicate of first request due to we moved to cybervisiontech fork of the repo.
Sorry for inconvenience. We will create pull requests from this fork in future.
All comments from first review was addressed except HttpClient and Mockito usage.


awholegunch:
Please see comment in PR #1 about package-info.java and spelling.

README.md
```diff
@@ -1,4 +1,26 @@
-cdap-ingest
-===========
+CDAP Ingest Project
+===================
 
-Various ways to ingest data into CDAP
+Various ways to ingest data into CDAP. 
+
+For more details, please, look to the README.md files in submodules.
```
awholegunch: For details, please see the README.md file in each submodule.

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
+stream-client-java
+==================
+
+Stream Client Java API for managing streams via external custom java applications.
```
awholegunch: The Stream Client Java API is for managing Streams via external custom Java applications.

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+## supported actions
```
awholegunch: ## Supported Actions

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+ - create Stream with specified <stream-id>;
```
awholegunch: `- create a Stream with a specified <stream-id>;`

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ - update TTL for Stream by specified <stream-id>;
```
awholegunch: `- update TTL for an exiting Stream with a specified <stream-id>;`

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ - get current Stream TTL by specified <stream-id>;
```
awholegunch: `- retrieve the current Stream TTL for a specified <stream-id>;`

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ - truncate Stream by specified <stream-id> (this means the deletion of all events that were written to the Stream);
```
awholegunch: - truncate an existing Stream (the deletion of all events that were written to the Stream);

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ - write event to the Stream by specified <stream-id>;
```
awholegunch: - write an event to an existing Stream; and

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ - send File to the Stream by specified <stream-id>.
```
awholegunch: - send a File to an existing Stream.

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+
+
+## usage
```
awholegunch: `## Usage [Note: write all titles with initial-capital case]`

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+ For the start to use Stream Client Java API, you need to include maven dependency to your project pom file:
```
awholegunch: To use the Stream Client Java API, include this Maven dependency in your project's `pom.xml` file:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+ <dependency>
+  <groupId>co.cask.cdap</groupId>
+  <artifactId>stream-client-java</artifactId>
+  <version>1.0-SNAPSHOT</version>
+ </dependency>
+ 
+## example
+   
+ Create StreamClient instance with mandatory fields 'host' and 'port' of Getaway server.
```
awholegunch: Create a StreamClient instance, specifying the fields 'host' and 'port' of the gateway server.

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ <dependency>
+  <groupId>co.cask.cdap</groupId>
+  <artifactId>stream-client-java</artifactId>
+  <version>1.0-SNAPSHOT</version>
+ </dependency>
+ 
+## example
+   
+ Create StreamClient instance with mandatory fields 'host' and 'port' of Getaway server. 
+ Optional configurations could be set as default:
```
awholegunch: Optional configurations that can be set (and their default values):

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+  
+  - ssl: false (use HTTP protocol) 
+  - writer pool size: '10' (max thread pool size for write events to the Stream)
+  - version : 'v2' (Getaway server version, used as a part of base uri [http://localhost:10000/v2/...])
```
awholegunch: ``-  Version : 'v2' (Gateway server version, used as a part of the base URI [http(s)://localhost:10000/v2/...])

stream-client/stream-client-java/README.md
awholegunch: [These Configurations should be either all caps (SSL) or initial caps (Writer Pool Size)...]

awholegunch: Either that or change "writer pool size" to "writerPoolSize" to match code...

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+## example
+   
+ Create StreamClient instance with mandatory fields 'host' and 'port' of Getaway server. 
+ Optional configurations could be set as default:
+  
+  - ssl: false (use HTTP protocol) 
+  - writer pool size: '10' (max thread pool size for write events to the Stream)
+  - version : 'v2' (Getaway server version, used as a part of base uri [http://localhost:10000/v2/...])  
+  - authToken: null (Need to be specify for authenticate client requests).
```
awholegunch: `- authToken: null (Need to specify to authenticate client requests)` [note no trailing period]

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+## example
+   
+ Create StreamClient instance with mandatory fields 'host' and 'port' of Getaway server. 
+ Optional configurations could be set as default:
+  
+  - ssl: false (use HTTP protocol) 
+  - writer pool size: '10' (max thread pool size for write events to the Stream)
+  - version : 'v2' (Getaway server version, used as a part of base uri [http://localhost:10000/v2/...])  
+  - authToken: null (Need to be specify for authenticate client requests). 
+  - apiKey: null (Need to be specify for authenticate client requests using SSL)
```
awholegunch: `- apiKey: null (Need to specify to authenticate client requests using SSL)` [Note is "using SSL" required for "authToken"? If so, please add to the above...]

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+ ```
+   StreamClient streamClient = new RestStreamClient.Builder("localhost", 10000).build();
+ ```
+      
+ or also could be specified as the builder parameters:
```
awholegunch: or specified using the builder parameters:

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,180 @@
...
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  public CloseableHttpResponse execute(HttpRequestBase request) throws IOException {
+    if (StringUtils.isNotEmpty(config.getAuthToken())) {
+      request.setHeader(HttpHeaders.AUTHORIZATION, AUTHENTICATION_HEADER_PREFIX_BEARER + config.getAuthToken());
+    }
+    if (StringUtils.isNotEmpty(config.getApiKey())) {
+      request.setHeader(CONTINUUITY_API_KEY_HEADER_NAME, config.getApiKey());
+    }
+    LOG.debug("Execute Http Request: " + request);
```
awholegunch: When using LOG messages, always write them as `LOG.debug("Execute Http Request: {}", request);` so that they will be optimized for different LOG levels.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,180 @@
...
+        throw new ServerErrorException("Internal server exception during operation process.",
+                                       HttpStatus.SC_INTERNAL_SERVER_ERROR);
+      case HttpStatus.SC_NOT_IMPLEMENTED:
+      default:
+        throw new NotSupportedException("Operation is not supported");
+    }
+  }
+
+  /**
+   * Utility method for convert {@link org.apache.http.HttpEntity} http entity content to JsonObject
```
awholegunch: Utility method for converting {@link org.apache.http.HttpEntity} HTTP entity content to JsonObject.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,180 @@
...
+   *
+   * @param httpEntity {@link org.apache.http.HttpEntity}
+   * @return {@link JsonObject} generated from input content stream
+   * @throws IOException in case if entity content is not available
```
awholegunch: @throws IOException if entity content is not available

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+   StreamClient streamClient = new RestStreamClient.Builder("localhost", 10000)
+         .apiKey("apiKey")
+         .authToken("token")
+         .ssl(false)
+         .version("v2")
+         .writerPoolSize(10)
+         .build();
+ ```
+ 
+ Create new Stream with new *stream-id*
```
awholegunch: Create a new Stream with the _stream-id_ "newStreamName":

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+   streamClient.create("newStreamName");
+ ```
+      
+ Notes:
+ 
+  - The <stream-id> should only contain ASCII letters, digits and hyphens.
+  - If the Stream already exists, no error is returned, and the existing Stream remains in place.
+     
+ 
+ Update TTL for Stream by *stream-id*, TTL is long value
```
awholegunch: Update TTL for the Stream _stream-id_; TTL is a long value:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+ ```
+   streamClient.setTTL("streamName", newTTL);
+ ```
+ 
+ Get current TTL value by *stream-id*
```
awholegunch: Get the current TTL value for the Stream _stream-id_:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ Notes:
+ 
+  - The <stream-id> should only contain ASCII letters, digits and hyphens.
+  - If the Stream already exists, no error is returned, and the existing Stream remains in place.
+     
+ 
+ Update TTL for Stream by *stream-id*, TTL is long value
+ 
+ ```
+   streamClient.setTTL("streamName", newTTL);
```
awholegunch, line 73: Shouldn't "streamName" be the same for all these examples? The "createWriter" example is using a variable and not a string.

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ ```
+ 
+ Get current TTL value by *stream-id*
+ 
+ ```  
+   long ttl = streamClient.getTTL("streamName");  
+ ```
+ 
+ Create StreamWriter instance for writing events to the Stream
```
awholegunch: Create a `StreamWriter` instance for writing events to the Stream _newStreamName_:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+ ```
+   StreamWriter streamWriter = streamClient.createWriter(streamName);
```
awholegunch: StreamWriter streamWriter = streamClient.createWriter("newStreamName"); [is this what this should be?]

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ ```
+     
+ For write new event to the stream, you could use on of existing 5 methods from the StreamWriter interface:
```
awholegunch: To write new events to the Stream, you can use any of these five methods in the `StreamWriter` interface:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+ ``` 
+   ListenableFuture<Void> write(String str, Charset charset);
+   ListenableFuture<Void> write(String str, Charset charset, Map<String, String> headers);
+   ListenableFuture<Void> write(ByteBuffer buffer);
+   ListenableFuture<Void> write(ByteBuffer buffer, Map<String, String> headers);
+   ListenableFuture<Void> send(File file, MediaType type);
+ ```
+ 
+ for example:
```
awholegunch: Example:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+ ```
+   streamWriter.write("New log event", Charsets.UTF_8).get();
+ ```
+   
+ For truncate Stream by *stream-id*, use
```
awholegunch: To truncate the Stream _newStreamName_, use:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+ ```
+   streamClient.truncate("streamName");
+ ```
+   
+ When you already does not need to use created clients, release all unused resources by calling methods
```
awholegunch: When you are finished, release all resources by calling these two methods, in this order:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ 
+ ```  
+   streamWriter.close();
+   streamClient.close();  
+ ```
+
+## additional notes
+ 
+ All methods from the StreamClient and StreamWriter throw Exceptions using response code analysis from
```
awholegunch: All methods from the `StreamClient` and `StreamWriter` throw exceptions using response code analysis from the gateway server. These exceptions help determine if the request was processed successfully or not.

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+   streamWriter.close();
+   streamClient.close();  
+ ```
+
+## additional notes
+ 
+ All methods from the StreamClient and StreamWriter throw Exceptions using response code analysis from 
+ the Getaway server. This exceptions help to recognizes is the request processed successfully or not.
+  
+ So, in the case of **200 OK** no exception will be thrown;
```
awholegunch: In the case of a **200 OK** response, no exception will be thrown; other cases will throw these exceptions:

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,136 @@
...
+ In other cases will be thrown following exceptions:
+  
+  - **400 Bad Request**
```
awholegunch: - **400 Bad Request**: _javax.ws.rs.BadRequestException_ [common format for this table]

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Utility method for analysis http response status code and throw appropriate Java API Exception
+   *
+   * @param response {@link HttpResponse} http response
+   */
+  public static void responseCodeAnalysis(HttpResponse response) {
+    int code = response.getStatusLine().getStatusCode();
+    switch (code) {
+      case HttpStatus.SC_OK:
+        LOG.info("Success operation result code");
```
chtyim: Should be debug, not info.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -0,0 +1,74 @@
...
+
+package co.cask.cdap.client;
+
+import co.cask.cdap.client.exception.NotFoundException;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+/**
+ * The client interface to interact with services provided by stream endpoint.
```
awholegunch: the Stream [common to all files]

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,91 @@
...
+import co.cask.cdap.client.rest.RestStreamClient;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import org.apache.log4j.BasicConfigurator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * As Example of usage StreamClient API
```
awholegunch: An example of using the StreamClient API.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public class Main {
+  private static final Logger LOG = LoggerFactory.getLogger(Main.class);
+
+  public static void main(String[] args) {
+    BasicConfigurator.configure();
+
+    final String streamName = "test";
+
+    // Create StreamClient Instance with mandatory fields 'host' and 'port'. Optional configurations will be set as
```
awholegunch: // Create StreamClient instance with mandatory fields 'host' and 'port'. 
// Optional configurations will be set as:

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Utility method for convert {@link org.apache.http.HttpEntity} http entity content to JsonObject
+   *
+   * @param httpEntity {@link org.apache.http.HttpEntity}
+   * @return {@link JsonObject} generated from input content stream
+   * @throws IOException in case if entity content is not available
+   */
+  public static JsonObject getEntityAsJsonObject(HttpEntity httpEntity) throws IOException {
+    JsonObject result;
+    if (httpEntity != null && httpEntity.getContent() != null) {
```
chtyim: Prefer to revert the case and throw earlier. That would make the code less indented and easier to read.

``` java
if (httpEntity == null || httpEntity.getContent() == null) {
  throw ...
}
ContentType contentType = ....
...
```

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,91 @@
...
+    // default: protocol : 'http', writer pool size: '10', version : 'v2'.
```
awholegunch: // defaults: protocol: 'http', writer pool size: 10, version : 'v2'.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,91 @@
...
+      LOG.info("Get TTL for {} stream: {}", currentTTL, streamName);
+      long newTTL = 10009907;
+
+      // Update TTL value for Stream by id <streamName>
+      streamClient.setTTL(streamName, newTTL);
+      LOG.info("Set new TTL for {} stream: {}", newTTL, streamName);
+
+      // Get current Stream TTL value by id <streamName> after updating for compare
+      currentTTL = streamClient.getTTL(streamName);
+      LOG.info("Is TTL was updated successfully? {}", currentTTL == newTTL ? "YES" : "NO");
```
awholegunch: LOG.info("Was TTL updated successfully? {}", currentTTL == newTTL ? "YES" : "NO");

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,91 @@
...
+
+      // Create StreamWriter Instance
+      StreamWriter streamWriter = streamClient.createWriter(streamName);
+
+      String log = "95.211.139.1 - - [09/Apr/2012:08:40:43 -0400] \"GET /NoteBook/ HTTP/1.0\" 201 809 \"-\" " +
```
awholegunch: String log = "192.0.2.0 - - [09/Apr/2012:08:40:43 -0400] \"GET /NoteBook/ HTTP/1.0\" 201 809 \"-\" " +

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,91 @@
...
+        "\"OpenAcoon v4.10.5 (www.openacoon.de)\"";
```
awholegunch: "\"Example v0.0.0 (www.example.org)\"";

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,91 @@
...
+
+      // Update TTL value for Stream by id <streamName>
+      streamClient.setTTL(streamName, newTTL);
+      LOG.info("Set new TTL for {} stream: {}", newTTL, streamName);
+
+      // Get current Stream TTL value by id <streamName> after updating for compare
+      currentTTL = streamClient.getTTL(streamName);
+      LOG.info("Is TTL was updated successfully? {}", currentTTL == newTTL ? "YES" : "NO");
+
+      // Create StreamWriter Instance
```
awholegunch: Below I have modified the example IP address and domain to use the ranges reserved for documentation.

stream-client/stream-client-java/README.md
amakogoncv: This is should be already existing Stream, so I think, will be better to use "streamName".

stream-client/stream-client-java/README.md
amakogoncv: > > "in this order" could be superfluous, because order doesn't matter in this case.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/example/Main.java
amakogoncv: Clear, thank you for all remarks above.

stream-client/stream-client-java/README.md
```diff
@@ -0,0 +1,127 @@
...
+   streamClient.create("newStreamName");
+ ```
+      
+ Notes:
+ 
+  - The <stream-id> should only contain ASCII letters, digits and hyphens.
+  - If the Stream already exists, no error is returned, and the existing Stream remains in place.
+     
+ 
+ Update TTL for the Stream *stream-id*; TTL is a long value:
```
awholegunch: Instead of "the Stream _stream-id_;", what about "the Stream _streamName_;" ?
 _stream-id_ is the name of the parameter rather than the Stream.

stream-client/stream-client-java/README.md
awholegunch: Does the StreamWriter contain a reference to the StreamClient, or vice-versa? Does one have to be disposed of first to prevent a resource lock?
JJ

amakogoncv: No, StreamWriter doesn't contain any references to the resources of StreamClient, StreamWriter just use the same configuration parameters (host, port, protocol etc.). StreamWriter has it own connection manager and connection client, so it is not matter in what order you will close these clients. For example, could be cases when you already doesn't need one of them, but still uses second one. This two objects use independent resources.

pom.xml
```diff
@@ -0,0 +1,250 @@
...
+        <version>${guava.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+        <version>${slf4j.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-log4j12</artifactId>
```
chtyim: Don't use log4j. Use logback instead. You can use `log4j-over-slf4j` to bridge log4j API into slf4j appender. You'll need to exclude log4j dependency from httpclient to avoid API conflict.

pom.xml
```diff
@@ -0,0 +1,250 @@
...
+        <artifactId>gson</artifactId>
+        <version>${gson.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>javax.ws.rs</groupId>
+        <artifactId>javax.ws.rs-api</artifactId>
+        <version>${rs-api.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.glassfish.jersey.core</groupId>
```
chtyim, line 114: What is this for?

stream-client/stream-client-java-example/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,92 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * An example of using the StreamClient API.
+ */
+public class Main {
+  private static final Logger LOG = LoggerFactory.getLogger(Main.class);
+
+  public static void main(String[] args) {
+    BasicConfigurator.configure();
```
chtyim: As mentioned above. Don't use log4j.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
```diff
@@ -0,0 +1,205 @@
...
+    private final String host;
+
+    //optional
+    private String authToken;
+    private String apiKey;
+    private boolean ssl = false;
+    private int writerPoolSize = DEFAULT_WRITER_POOL_SIZE;
+    private String version = DEFAULT_VERSION;
+
+    public Builder(String host, int port) {
```
chtyim: A better builder instantiation pattern would be having the Builder constructor `private` and have a static `builder()` method in the outer class. The usage pattern is like:

``` java
RestStreamClient.builder(host, port).create();
```

stream-client/stream-client-java-example/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,92 @@
...
+      .build();
+
+    try {
+      // Create Stream by id <streamName>
+      streamClient.create(streamName);
+
+      // Get current Stream TTL value by id <streamName>
+      long currentTTL = streamClient.getTTL(streamName);
+      LOG.info("Get TTL for {} stream: {}", currentTTL, streamName);
+      long newTTL = 10009907;
```
chtyim: This is a very interesting TTL value (115.855405 days)? How do you come up with this?

stream-client/stream-client-java-example/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,92 @@
...
+      streamClient.setTTL(streamName, newTTL);
+      LOG.info("Set new TTL for {} stream: {}", newTTL, streamName);
+
+      // Get current Stream TTL value by id <streamName> after updating for compare
+      currentTTL = streamClient.getTTL(streamName);
+      LOG.info("Was TTL updated successfully? {}", currentTTL == newTTL ? "YES" : "NO");
+
+      // Create StreamWriter Instance
+      StreamWriter streamWriter = streamClient.createWriter(streamName);
+
```
chtyim, line 45: Should be a `try-finally` block to close the StreamWriter created.

stream-client/stream-client-java-example/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -0,0 +1,92 @@
...
+  public static void main(String[] args) {
+    BasicConfigurator.configure();
+
+    final String streamName = "test";
+
+    // Create StreamClient instance with mandatory fields 'host' and 'port'.
+    // Optional configurations will be set as:
+    // defaults: protocol : 'http', writerPoolSize: '10', version : 'v2'.
+    StreamClient streamClient = new RestStreamClient.Builder("localhost", 10000)
```
chtyim: Should close the StreamClient in the `try-finally` block.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamWriter.java
```diff
@@ -0,0 +1,99 @@
...
+  ListenableFuture<Void> send(File file, MediaType type);
+
+  /**
+   * Closes a {@link org.apache.http.impl.client.CloseableHttpClient} instance and a
+   * {@link com.google.common.util.concurrent.ListeningExecutorService} Executor thread pool
+   * for releasing all unused resource associated with them.
+   *
+   * @throws IOException if an I/O error occurs during close process
+   */
+  void close() throws IOException;
```
chtyim, line 99: Annotate with `@Override`

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  StreamWriter createWriter(String stream) throws NotFoundException, IOException;
+
+  /**
+   * Closes a {@link org.apache.http.impl.client.CloseableHttpClient} instance for releasing all unused resource
+   * associated with it.
+   *
+   * @throws IOException if an I/O error occurs during close process
+   */
+  void close() throws IOException;
```
chtyim, line 76: Annotate with `@Override`.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/exception/NotFoundException.java
```diff
@@ -0,0 +1,46 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.client.exception;
+
+/**
+ * Thrown when an element is not found
+ */
+public class NotFoundException extends Exception {
```
chtyim: I think we can use the `javax.ws.rs.NotFoundException` directly. I know that one is a RuntimeException, but it's ok. Also, the methods in `StreamClient` is better modified to not throwing exception. Instead, exceptions should be reflected as the future failure cause.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+  private static final String PROTOCOL_POSTFIX = "://";
+  private static final String COLON = ":";
+
+  private final RestClientConnectionConfig config;
+  private final String baseUrl;
+  private final CloseableHttpClient httpClient;
+
+  public RestClient(RestClientConnectionConfig config, HttpClientConnectionManager connectionManager) {
+    this.config = config;
+    this.baseUrl = config.isSsl() ? HTTPS_PROTOCOL : HTTP_PROTOCOL + PROTOCOL_POSTFIX + config.getHost() +
```
chtyim: Use `String.format` instead. It'll make the code more readable. E.g.

``` java
String.format("%s://%s:%d/%s", protocol, host, port, path);
```

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * Utility method for converting {@link org.apache.http.HttpEntity} HTTP entity content to JsonObject.
+   *
+   * @param httpEntity {@link org.apache.http.HttpEntity}
+   * @return {@link JsonObject} generated from input content stream
+   * @throws IOException if entity content is not available
+   */
+  public static JsonObject getEntityAsJsonObject(HttpEntity httpEntity) throws IOException {
```
chtyim: Simply call this `toJsonObject`

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * Utility method for converting {@link org.apache.http.HttpEntity} HTTP entity content to String.
+   *
+   * @param httpEntity {@link org.apache.http.HttpEntity}
+   * @return {@link String} generated from input content stream
+   * @throws IOException if HTTP entity is not available
+   */
+  public static String getEntityAsString(HttpEntity httpEntity) throws IOException {
```
chtyim: Simply call this `toString`

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+    Charset charset;
+    ContentType contentType = ContentType.getOrDefault(httpEntity);
+    if (contentType != null && contentType.getCharset() != null) {
+      charset = contentType.getCharset();
+    } else {
+      charset = Charsets.UTF_8;
+    }
+    Reader reader = new InputStreamReader(httpEntity.getContent(), charset);
+    try {
+      content = CharStreams.toString(reader);
```
chtyim: Simply return from here, then you don't need the extra `content` local variable.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+   * @param httpEntity {@link org.apache.http.HttpEntity}
+   * @return {@link String} generated from input content stream
+   * @throws IOException if HTTP entity is not available
+   */
+  public static String getEntityAsString(HttpEntity httpEntity) throws IOException {
+    String content = null;
+    if (httpEntity == null || httpEntity.getContent() == null) {
+      throw new IOException("Empty HttpEntity is received.");
+    }
+    Charset charset;
```
chtyim: Have `charset` default to `Charsets.UTF_8`, then you don't need the else clause below, making the code cleaner.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+    ContentType contentType = ContentType.getOrDefault(httpEntity);
+    if (contentType != null && contentType.getCharset() != null) {
```
chtyim, line 153: `getOrDefault` never returns `null`, no need to check for that.

chtyim, line 153: In fact, you should call `ContentType.get()` as you don't want it to get defaulted (default is ISO8859-1, not UTF-8).

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+   * Utility method for converting {@link org.apache.http.HttpEntity} HTTP entity content to JsonObject.
+   *
+   * @param httpEntity {@link org.apache.http.HttpEntity}
+   * @return {@link JsonObject} generated from input content stream
+   * @throws IOException if entity content is not available
+   */
+  public static JsonObject getEntityAsJsonObject(HttpEntity httpEntity) throws IOException {
+    JsonObject result;
+    String content = getEntityAsString(httpEntity);
+    if (StringUtils.isNotEmpty(content)) {
```
chtyim: Revert the condition and throw earlier.

``` java
if (StringUtils.isEmpty(content)) {
  throw new IOException(...);
}
return new JsonParser().parse(content).getAsJsonObject();
```

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+    return content;
+  }
+
+  /**
+   * Method for releasing unused resources
+   *
+   * @throws IOException if an I/O error occurs
+   */
+  public void close() throws IOException {
+    if (httpClient != null) {
```
chtyim: the `httpClient` never be `null` as it's a `final` field get initialized in constructor.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClient.java
```diff
@@ -0,0 +1,190 @@
...
+      httpClient.close();
+    }
+  }
+
+  /**
+   * @return the base URL of Rest Service API
+   */
+  public String getBaseUrl() {
```
chtyim: Use `URL` class to represent url.

chtyim: Also, the method should be named `getBaseURL()`.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClientConnectionConfig.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public String getHost() {
+    return host;
+  }
+
+  public String getVersion() {
+    return version;
+  }
+
+  public boolean isSsl() {
```
chtyim: SSL should be all caps. `isSSL()`.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestClientConnectionConfig.java
```diff
@@ -0,0 +1,64 @@
...
+    return ssl;
+  }
+
+  public String getApiKey() {
```
chtyim: API should be all caps. In Java, acronyms should be all caps when used in method name.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
```diff
@@ -0,0 +1,205 @@
...
+import java.io.IOException;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * Stream client implementation used REST Api for stream management
+ */
+public class RestStreamClient implements StreamClient {
+  private static final Logger LOG = LoggerFactory.getLogger(RestStreamClient.class);
+
+  public static final String DEFAULT_VERSION = "v2";
```
chtyim: Why these three fields are public? If just for testing purpose, make them package private so that only test cases in the same package can see them.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
```diff
@@ -0,0 +1,205 @@
...
+
+  private static final Gson GSON = new Gson();
+
+  private final RestClientConnectionConfig config;
+  private final int writerPoolSize;
+  private final RestClient restClient;
+
+  @Override
+  public void create(String stream) throws IOException {
+    HttpPut putRequest = new HttpPut(restClient.getBaseUrl() + String.format("streams/%s", stream));
```
chtyim: When constructing new URL (URI) from base, better use `URI` to represents the base and use `URI.resolve` to resolve it full path.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/exception/NotFoundException.java
chtyim: Forget about my comment on the future part, as I overlooked that method on the StreamClient are not returning future. But I think using `javax.ws.rs.NotFoundException` is ok.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
```diff
@@ -0,0 +1,205 @@
...
+      httpResponse.close();
+    }
+  }
+
+  @Override
+  public StreamWriter createWriter(String stream) throws NotFoundException, IOException {
+    //get the Stream TTL for check does the requested Stream exist
+    long ttl = getTTL(stream);
+    LOG.debug("The Stream with id {} exists. Got the current Stream TTL value {} successfully.", stream, ttl);
+    PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager();
```
chtyim, line 125: So each time when the user call the `createWriter` method, a new pool manager will be created? Shouldn't they share the same connection pool?

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
```diff
@@ -0,0 +1,205 @@
...
+      this.version = version;
+      return this;
+    }
+
+    public RestStreamClient build() {
+      return new RestStreamClient(this);
+    }
+  }
+
+  private RestStreamClient(Builder builder) {
```
chtyim: Move constructor before any method.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
```diff
@@ -0,0 +1,205 @@
...
+    writerPoolSize = builder.writerPoolSize;
+    config = new RestClientConnectionConfig(builder.host, builder.port, builder.authToken, builder.apiKey,
+                                            builder.ssl, builder.version);
+    restClient = new RestClient(config, new PoolingHttpClientConnectionManager());
+  }
+
+  public RestClient getRestClient() {
+    return restClient;
```
chtyim: Isn't this internal, why returning it?

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  public int getWriterPoolSize() {
+    return writerPoolSize;
```
chtyim: Why need to expose this?

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
```diff
@@ -0,0 +1,145 @@
...
+
+  @Override
+  public ListenableFuture<Void> write(String str, Charset charset) {
+    return write(str, charset, null);
+  }
+
+  @Override
+  public ListenableFuture<Void> write(String str, Charset charset, Map<String, String> headers) throws
+    IllegalArgumentException {
+    if (StringUtils.isNotEmpty(str)) {
```
chtyim: Use `Preconditions.checkArgument`.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
```diff
@@ -0,0 +1,145 @@
...
+
+  public RestStreamWriter(RestClient restClient, int writerPoolSize, String streamName) {
+    this.restClient = restClient;
+    this.streamName = streamName;
+    this.pool = MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(writerPoolSize));
+  }
+
+  @Override
+  public ListenableFuture<Void> write(String str, Charset charset) {
+    return write(str, charset, null);
```
chtyim: Better use an empty map than `null` to make logic simpler (`ImmutableMap.of()`)

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
```diff
@@ -0,0 +1,145 @@
...
+
+  @Override
+  public ListenableFuture<Void> write(ByteBuffer buffer) {
+    return write(buffer, null);
+  }
+
+  @Override
+  public ListenableFuture<Void> write(ByteBuffer buffer, Map<String, String> headers) throws IllegalArgumentException {
+    if (buffer != null) {
+      byte[] bytes = new byte[buffer.remaining()];
```
chtyim, line 84: Only creates a new array if the given `ByteBuffer` is not back by array.

``` java
HttpEntity content;
if (buffer.hasArray()) {
  content = new ByteArrayEntity(buffer.array(), buffer.arrayOffset() + buffer.position(), buffer.remaining());
} else {
  // The existing logic.
}
```

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
```diff
@@ -0,0 +1,145 @@
...
+        } finally {
+          response.close();
+        }
+        return null;
+      }
+    });
+  }
+
+  public void close() throws IOException {
+    if (pool != null) {
```
chtyim: cannot be null.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
```diff
@@ -0,0 +1,145 @@
...
+      pool.shutdown();
+    }
+    if (restClient != null) {
```
chtyim: cannot be null.

pom.xml
amakogoncv, line 114: javax.ws.rs
For use javax.ws.rs Exceptions;

commons-lang
For clean code, but it is not necessary

stream-client/stream-client-java-example/src/main/java/co/cask/cdap/client/example/Main.java
amakogoncv: It is just a random value for test in the local env, nothing more. But I will change it to no one is confused :)

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
amakogoncv: No, I think it is not need the same as `public RestClient getRestClient()` too. I will remove both these getters.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamClient.java
amakogoncv, line 125: I believe, they should not. `StreamWriter` use own ConnectionManager to increase performance and have ability to set pool size as much as allows getaway server. But `StreamClient` doesn't need such big thread pool, because it will be used much less.

pom.xml
chtyim, line 114: I mean for the jersey-common. Where do we use it?

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
```diff
@@ -0,0 +1,141 @@
...
+          RestClient.responseCodeAnalysis(response);
+        } finally {
+          response.close();
+        }
+        return null;
+      }
+    });
+  }
+
+  public void close() throws IOException {
```
chtyim, line 126: Missing `@Override`

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
```diff
@@ -0,0 +1,141 @@
...
+
+  public RestClient getRestClient() {
+    return restClient;
+  }
+
+  public String getStreamName() {
+    return streamName;
+  }
+
+  public ListeningExecutorService getPool() {
```
chtyim: Why need to expose the internal pool?

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
```diff
@@ -0,0 +1,141 @@
...
+      }
+    });
+  }
+
+  public void close() throws IOException {
+    pool.shutdown();
+    restClient.close();
+  }
+
+  public RestClient getRestClient() {
```
chtyim: Why need to expose?

pom.xml
amakogoncv, line 114: Sorry for misunderstanding. For throw javax.ws.rs exceptions. If this dependency is missing, we get java.lang.ClassNotFoundException: org.glassfish.jersey.internal.RuntimeDelegateImpl instead of expected javax.ws.rs Exception (BadRequestException, NotFoundException etc.).

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/rest/RestStreamWriter.java
amakogoncv: It makes sense to remove these two getters, because these are internal  fields for use only inside `StreamWriter` and `StreamClient`

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -0,0 +1,78 @@
...
+ * The client interface to interact with services provided by the Stream endpoint.
+ */
+public interface StreamClient extends Closeable {
+  /**
+   * Creates the Stream with the given name.
+   *
+   * @param stream String value of the Stream id
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  void create(String stream) throws IOException;
```
sreevatsanraman, line 32: What will happen if stream already exists or Stream Name is not valid? We should throw appropriate exceptions in that case.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Set the Time-To-Live (TTL) property of the given Stream.
+   *
+   * @param stream Name of the Stream
+   * @param ttl    TTL in seconds
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  void setTTL(String stream, long ttl) throws IOException;
```
sreevatsanraman, line 41: What happens when Stream doesn't exist? Need to throw appropriate exception

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Retrieves the Time-To-Live (TTL) property of the given Stream.
+   *
+   * @param stream Name of the Stream
+   * @return Current TTL of the Stream in seconds
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  long getTTL(String stream) throws IOException;
```
sreevatsanraman, line 50: Same here.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Truncates all existing events in the give Stream.
+   *
+   * @param stream Name of the Stream
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  void truncate(String stream) throws IOException;
```
sreevatsanraman, line 58: Same here.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Creates a {@link StreamWriter} instance for writing events to the given Stream.
+   *
+   * @param stream Name of the Stream
+   * @return An instance of {@link StreamWriter} that is ready for writing events to the Stream
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  StreamWriter createWriter(String stream) throws IOException, URISyntaxException;
```
sreevatsanraman: Why expose URISyntaxException? Should be something higher level and not expose the underlying implementation details

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -0,0 +1,78 @@
...
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  void truncate(String stream) throws IOException;
+
+  /**
+   * Creates a {@link StreamWriter} instance for writing events to the given Stream.
+   *
+   * @param stream Name of the Stream
+   * @return An instance of {@link StreamWriter} that is ready for writing events to the Stream
+   * @throws IOException in case of a problem or the connection was aborted
```
sreevatsanraman, line 65: Javadocs should reflect all the exceptions thrown

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
nitinmotgi, line 50: Aren't IOException the right thing to be thrown ?

---

"I’ve learned that I still have a lot to learn. I’ve learned that people will forget what you said, people will forget what you did, but people will never forget how you made them feel" - Maya Angelou

Random auto-corrects and typos are my special gift to you. When I forward they are from others.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
nitinmotgi: aren't these the right things ?

---

"I’ve learned that I still have a lot to learn. I’ve learned that people will forget what you said, people will forget what you did, but people will never forget how you made them feel" - Maya Angelou

Random auto-corrects and typos are my special gift to you. When I forward they are from others.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
amakogoncv, line 32: Exceptions are thrown according to the response status code from the gateway server. According to REST API documentation: "If the Stream already exists, no error is returned, and the existing Stream remains in place." So in this case exception will not thrown.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
amakogoncv, line 41: NotFoundException will be thrown.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
amakogoncv, line 50: NotFoundException will be thrown, according to discussions with Terence.

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
amakogoncv, line 58: The same answer :)

stream-client/stream-client-java/src/main/java/co/cask/cdap/client/StreamClient.java
amakogoncv, line 65: ok, I will fix it.

sreevatsanraman:
There are four issues that needs to be addressed in this PR
a) Add guava dependency to stream-client/stream-client-java/pom.xml  - I was not able to compile without this
b) Answer Terence's comment about jersey-dependency
c) Add instructions on how to build in README.md
d) When I tried to compile the unit-tests failed, need to make sure that the unit tests work

sreevatsanraman:
Spoke to Alina about the concerns raised. The jersey dependency is an open issue. We will resolve this after we close this PR. opened a jira to track - https://jira.continuuity.com/browse/CYB-67

sreevatsanraman:
LGTM :+1:

PR #4, opened 2014-08-22
CYB-4: [Stream Ingestion] Design and Implement Python API for ingesting into stream 

ashvayka:
Please note that this PR depends on PR 2 (CYB-15: [Stream Ingestion] Design and Implement Java API for ingesting into stream). Actual PR contains only from stream-client/stream-client-python


stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
+# CDAP ingest Python library
```
awholegunch: CDAP Ingest Python library

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+Stream Client Python API for managing streams via external custom Python applications.
```
awholegunch: managing Streams via external Python applications.

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+- create a Stream with a specified <stream-id>;
+- update TTL for an existing Stream with a specified <stream-id>;
+- retrieve the current Stream TTL for a specified <stream-id>;
+- truncate an existing Stream (the deletion of all events that were written to the Stream);
+- write an event to an existing Stream specified by <stream-id>;
+- send a File to an existing Stream specififed by <stream-id>.
+
+## Usage
+
+ To use the Stream Client Python API, include this into your script:
```
awholegunch: To use the Stream Client Python API, include in your Python script:

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+import config
```
awholegunch: Put four spaces at the start of each line of code (to mark them as code blocks), and add a blank line before and after the "or" line to separate the two code block clearly.

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+import streamclient
+or
+from config import Config
+from streamclient import StreamClient
```
awholegunch: Reading your sample code below: it won't work if you do "import config". You'd need "config = config.Config()", which would redefine config and get really confusing. Why not just show the single import example ("from config import Config", etc.) as Pythonistas will understand what you are talking about?

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+   streamClient = StreamClient(config)
+ ```
+
+Config file structure:
+```
+[ServerConnection]
+hostname = kappac             - gateway hostname
+port = 10001                  - gateway port
+SSL = true                    - should be SSL used
```
awholegunch: if SSL is being used

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+ ```
+   config = Config.readFromFile('/path/to/config.ini')
+
+   streamClient = StreamClient(config)
+ ```
+
+Config file structure:
+```
+[ServerConnection]
```
awholegunch: I'm not sure what this line is for? Maybe it's just me...

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+   streamClient.create("newStreamName");
+ ```
+
+ Notes:
+
+  - The <stream-id> should only contain ASCII letters, digits and hyphens.
+  - If the Stream already exists, no error is returned, and the existing Stream remains in place.
+
+
+ Update TTL for the Stream *stream-id*; TTL is a long value:
```
awholegunch: Update TTL for the Stream "streamName";  _newTTL_ is a long value:

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+ ```
+   streamClient.setTTL("streamName", newTTL);
+ ```
+
+ Get the current TTL value for the Stream *stream-id*:
```
awholegunch: et the current TTL value for the Stream "stream-name":

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+ ```
+
+ To truncate the Stream *streamName*, use:
+
+ ```
+   streamClient.truncate("streamName");
+ ```
+
+ ### StreamPromise
+ StreamPromise's goal is to implement deffered code execution.
```
awholegunch: StreamPromise's goal is to implement deferred code execution.

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+ To handle successful an error handling you should create a handler for each case and set this handlers using ```onResponse``` method.
```
awholegunch: For error handling, create a handler for each case and set it using the `onResponse` method.

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+def onErrorHandler(httpResponse):
+    ...
+    parse response
+    ...
+
+streamPromise.onResponse(onOkResponse, onErrorResponse)
+```
+
+It's possible to not define an error handler. In that case successful handler would used in both cases.
```
awholegunch: I am have trouble parsing this. Do you mean:
"It's not required to define an error handler. If you don't specify an error handler, the success handler is used for both success and error conditions."

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Additional Notes
+
+ All methods from the ```StreamClient``` and ```StreamWriter``` throw exceptions using response code analysis from the 
+ gateway server. These exceptions help determine if the request was processed successfully or not.
+
+ In the case of a **200 OK** response, no exception will be thrown; other cases will throw the NoFoundException exception.
```
awholegunch, line 133: "NoFoundException": is that a good name for that exception? Why not raise urllib.error.URLError ?

stream-client/stream-client-python/config.ini
```diff
@@ -0,0 +1,5 @@
+[ServerConnection]
+hostname = kappac
+port = 10001
+SSL = true
+APIKey = something
```
awholegunch: Shouldn't this be blank?

stream-client/stream-client-python/config.ini
```diff
@@ -0,0 +1,5 @@
+[ServerConnection]
+hostname = kappac
```
awholegunch: Should this be localhost?

stream-client/stream-client-python/streamwriter.py
```diff
@@ -0,0 +1,221 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+from os import path
+from threading import Thread, Lock
+from types import FunctionType
+import locale
+import mimetypes
+from serviceconnector import *
```
awholegunch: Do you need to do import *? That's usually frowned upon...

stream-client/stream-client-python/streamwriter.py
```diff
@@ -0,0 +1,221 @@
...
+                self.__onOkHandler = self.__onErrorHandler = None
+        self.__handlersLock.release()
+
+    def onResponse(self, ok, error=None):
+        """
+        Sets up handlers for successful and error responses.
+
+        Keyword arguments:
+        ok -- Handler to be called in case of successful response.
+        error -- Handler to be calles in case of failed response.
```
awholegunch: Handler to be called in case of failed response.

stream-client/stream-client-python/streamwriter.py
```diff
@@ -0,0 +1,221 @@
...
+            ...
+            fooling around with response
+            ...
+        """
+        if not isinstance(ok, FunctionType) or (not None == error and not isinstance(error, FunctionType)):
+            raise TypeError('parameters should be functions')
+
+        self.__handlersLock.acquire()
+        self.__onOkHandler = ok
+        if None == error:
```
awholegunch: Is there a reason why this isn't "if error == None:"?

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,85 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import requests
+from config import Config
+
+class NoFoundException(Exception):
```
awholegunch: Python exception names usually end with "Error", even though they are derived from "Exception". What about calling this class "StreamClientError"? What about adding something like:

```
def __str__(self):
    return "Code: %s \nMessage: %s" % (self.__errorCode, self.__errorMsg)
```

so that the error can be easily printed.

stream-client/stream-client-python/streamclient.py
```diff
@@ -0,0 +1,112 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+from serviceconnector import *
```
awholegunch: import \* ?

stream-client/stream-client-python/test/runtests.py
```diff
@@ -0,0 +1,270 @@
...
+import unittest, httpretty
+
+import os,sys,inspect
+currentdir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+parentdir = os.path.dirname(currentdir)
+sys.path.insert(0,parentdir)
+
+from serviceconnector import NoFoundException
+from streamwriter import StreamWriter
+from streamclient import *
```
awholegunch: Let's be explicit in imports and not use import *.

stream-client/stream-client-python/README.md
kpovetkin: http://en.wikipedia.org/wiki/INI_file

stream-client/stream-client-python/README.md
kpovetkin, line 133: Names are given according to this doc:
https://wiki.continuuity.com/pages/viewpage.action?spaceKey=VEN&title=Data+Ingestion+Requirements

stream-client/stream-client-python/streamwriter.py
kpovetkin: It's a habit.

In JavaScript, C, C++ it would lead to logick faults:

if( variable_name = null )

stream-client/stream-client-python/serviceconnector.py
kpovetkin: Names are given according to this doc:
https://wiki.continuuity.com/pages/viewpage.action?spaceKey=VEN&title=Data+Ingestion+Requirements

stream-client/stream-client-python/README.md
awholegunch: What about then changing this to:
Config file structure in [INI format](http://en.wikipedia.org/wiki/INI_file):

stream-client/stream-client-python/README.md
awholegunch, line 133: Please review these with Terence or Sree; the thinking here is that we should follow Python (and each language's) conventions wherever possible.

stream-client/stream-client-python/config.ini
```diff
@@ -0,0 +1,5 @@
+[ServerConnection]
+hostname = local.host
+port = 10001
```
ParthGandhi: Why 10001?

stream-client/stream-client-python/config.py
```diff
@@ -0,0 +1,92 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+from configparser import ConfigParser
+
+class Config:
+    __host = 'localhost'
+    __port = 10000
+    __useSSL = False
+    __apiKey = ''
```
ParthGandhi: We shouldn't need API keys anywhere since we don't intend to support sandboxes for the coming releases.

stream-client/stream-client-python/config.py
```diff
@@ -0,0 +1,92 @@
...
+        self.__useSSL = ssl
+
+    def getAPIKey(self):
+        return self.__apiKey
+
+    def setAPIKey(self, apiKey):
+        self.__apiKey = apiKey
+
+    def readFromFile(filename):
+        newConfig = Config()
```
ParthGandhi: `config = Config()`

Just make sure that the name "config" (of the module) isn't reassigned because of this. I don't think it would be.

stream-client/stream-client-python/README.md
ParthGandhi, line 133: Agree with @awholegunch . Please bring this up with either @chtyim or @sreevatsanraman .

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import requests
+from config import Config
+
+class NoFoundErrorn(Exception):
```
ParthGandhi: `NotFoundError`?

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+        return self.__errorMsg
+
+    def __str__(self):
+        return "Code: %s \nMessage: %s" % (self.__errorCode, self.__errorMsg)
+
+class ConnectionErrorChecker:
+    __HTTP_OK = 200
+
+    def checkResponseErrors(self, httpResponse):
+        if not self.__HTTP_OK == httpResponse.status_code:
```
ParthGandhi: `if self.HTTP_OK is not httpResponse.status_code:`

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+
+class ServiceConnector:
+    __base_url = ''
+    __connectionConfig = None
+    __defaultHeaders = {
+        'Authorization': 'Bearer ',
+        'X-Continuuity-ApiKey': ''
+    }
+
+    def __init__(self, config = Config()):
```
ParthGandhi: Default arguments usually don't have spaces around the `=`.

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+        if not isinstance(config, Config):
+            raise TypeError('parameter should be of type Config')
+
+        self.__connectionConfig = config
+
+        if self.__connectionConfig.getSSL():
+            self.__base_url = 'https://'
```
ParthGandhi: Better to make this a local variable called `protocol` and then do a str.format using it later (as opposed to just appending to it).

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+        else:
+            self.__base_url = 'http://'
+
+        self.__base_url += '{0}:{1}'.format(
+            self.__connectionConfig.getHost(),
+            self.__connectionConfig.getPort()
+        )
+
+        self.__defaultHeaders['X-Continuuity-ApiKey'] = self.__connectionConfig.getAPIKey()
```
ParthGandhi: Not needed.

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def setAuthorizationToken(self, token):
+        self.__defaultHeaders['Authorization'] = 'Bearer ' + token
+
+    def request(self, method, uri, body = None, headers = None):
+        headersToSend = self.__defaultHeaders
+        url = '{0}{1}'.format(self.__base_url, uri)
+
+        if not None == headers:
```
ParthGandhi: `if headers is not None:`.

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+            headersToSend.update(headers)
+
+        return requests.request(
+            method,
+            url,
+            data = body,
```
ParthGandhi: Same here, no spaces around the "=".

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+            headers = headersToSend
+        )
+
+    def send(self, uri, fields = None, headers = None):
+        headersToSend = self.__defaultHeaders
+        url = '{0}{1}'.format(self.__base_url, uri)
+
+        if not None == headers:
```
ParthGandhi: `if headers is not None:`.

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+            headersToSend.update(headers)
+
+        return requests.post(
```
ParthGandhi: Splitting it onto multiple lines doesn't always help readability unless its a long method call, which it isn't in this case.

stream-client/stream-client-python/streamclient.py
```diff
@@ -0,0 +1,113 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+from config import Config
+from serviceconnector import ServiceConnector, ConnectionErrorChecker
+from streamwriter import StreamWriter
+import json
```
ParthGandhi: External modules must be imported before our classes.

stream-client/stream-client-python/streamclient.py
```diff
@@ -0,0 +1,113 @@
...
+
+class StreamClient(ConnectionErrorChecker):
+    __serviceConnector = None
+    __serviceConfig = None
+
+    __BASE_VER = '/v2'
```
ParthGandhi: Call this `GATEWAY_VERSION` to stay consistent with the rest of our codebase.

stream-client/stream-client-python/streamclient.py
```diff
@@ -0,0 +1,113 @@
...
+    __REQUEST_PLACEHOLDERS = {
```
ParthGandhi: You don't really need this, do you? Its being used only once and always in the same way.

stream-client/stream-client-python/streamclient.py
```diff
@@ -0,0 +1,113 @@
...
+    def create(self, stream):
+        """
+        Creates a stream with the given name.
+
+        Keyword arguments:
+        stream -- stream name to create
+        """
+        uri = self.__prepareUri('stream', data=stream)
+
+        self.__serviceConnector.request('PUT', uri)
```
ParthGandhi: Would this also need response error checking?

stream-client/stream-client-python/serviceconnector.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def __str__(self):
+        return "Code: %s \nMessage: %s" % (self.__errorCode, self.__errorMsg)
+
+class ConnectionErrorChecker:
+    __HTTP_OK = 200
+
+    def checkResponseErrors(self, httpResponse):
+        if not self.__HTTP_OK == httpResponse.status_code:
+            raise NoFoundErrorn(httpResponse.status_code, httpResponse.reason)
```
ParthGandhi: Do we also want to check for other return codes? e.g. a 401 implies you're missing an AuthToken. 

Or do you do that somewhere else? Can't seem to find it.

stream-client/stream-client-python/config.py
```diff
@@ -0,0 +1,92 @@
...
+                'ssl': 'SSL',
+                'apikey': 'APIKey'
+            }
+        }
+    }
+
+    def __init__(self):
+        pass
+
+    def getHost(self):
```
ParthGandhi: Prefer getters and setters using the `@property` annotation.

stream-client/stream-client-python/streamwriter.py
ParthGandhi: Do: `if error is None:`

stream-client/stream-client-python/streamwriter.py
```diff
@@ -0,0 +1,221 @@
...
+    __onErrorHandler = None
+    __handlersLock = None
+
+    def __init__(self, serviceConnector, uri, data):
+        """
+        Object constructor
+
+        Keyword arguments:
+        serviceConnector -- reference to connection pool to communicate with gatewway
+        uri -- REST URL part to perform request.  Example: '/v2/strems/myStream'
```
ParthGandhi: "/v2/streams/mystream"

stream-client/stream-client-python/streamwriter.py
```diff
@@ -0,0 +1,221 @@
...
+        self.__onOkHandler = ok
+        if None == error:
+            self.__onErrorHandler = ok
+        else:
+            self.__onErrorHandler = error
+
+        self.__handlersLock.release()
+        self.__responseCheckTarget()
+
+class StreamWriter:
```
ParthGandhi: I think both these classes should be in separate files, since neither is particularly small.

stream-client/stream-client-python/test/runtests.py
```diff
@@ -0,0 +1,273 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import unittest, httpretty
```
ParthGandhi: Import on separate lines.

stream-client/stream-client-python/test/runtests.py
```diff
@@ -0,0 +1,273 @@
...
+
+import os,sys,inspect
+currentdir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
```
ParthGandhi: Could you elaborate on what you're trying to do here?

stream-client/stream-client-python/test/runtests.py
```diff
@@ -0,0 +1,273 @@
...
+
+        def onResponse(response):
+            self.exit_code = response.status_code
+
+        q = sw.write(self.messageToWrite)
+        q.onResponse(onResponse)
+
+        self.assertEqual(self.exit_code, 200)
+
+if '__main__' == __name__:
```
ParthGandhi: No spaces around "==".

stream-client/stream-client-python/test/some.log
```diff
@@ -0,0 +1,46 @@
+>>> Downloading 'ftp://ftp.spline.inf.fu-berlin.de/mirrors/gentoo/distfiles/push-1.6.tar.gz'
```
ParthGandhi: Mistakenly committed?

ParthGandhi:
Done with an initial review. Feel free to ask me if you have any specific questions. 

Some general notes:
1. This has been written in 3.x. As per @nitinmotgi and @wolf31o2 , we've decided to start by supporting 2.6 since that is still the most widely used version in Enterprise (sadly). 
2. Abstract classes should be done via https://docs.python.org/2/library/abc.html, unless theres a good reason not to. 
3. Please get acquainted with pep-8 (http://legacy.python.org/dev/peps/pep-0008). It is the style guide for Python. You may also use https://github.com/jcrocholl/pep8 to help you find mistakes.

stream-client/stream-client-python/config.ini
```diff
@@ -0,0 +1,5 @@
+[ServerConnection]
```
ParthGandhi: Also, I'd prefer a JSON config over .ini. Thoughts?

stream-client/stream-client-python/serviceconnector.py
kpovetkin: awholegunch added a note 6 days ago
Python exception names usually end with "Error", even though they are derived from "Exception". What about calling this class "StreamClientError"?

stream-client/stream-client-python/test/runtests.py
kpovetkin: currentdir = os.path.dirname(
    os.path.abspath(inspect.getfile(inspect.currentframe())))
parentdir = os.path.dirname(currentdir)
sys.path.insert(0, parentdir)

Is necessary to make possible imports from parent directory.

stream-client/stream-client-python/test/runtests.py
kpovetkin: PEP8 says: "should be"

http://legacy.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

stream-client/stream-client-python/test/some.log
kpovetkin: No, it's not.
It's random log file for testing.

stream-client/stream-client-python/config.ini
kpovetkin: Changed.

stream-client/stream-client-python/serviceconnector.py
ParthGandhi: `StreamClientError` is too generic. I'd prefer something more specific.

ParthGandhi: Perhaps `StreamNotFoundError` is more meaningful.

stream-client/stream-client-python/test/runtests.py
ParthGandhi: Yes, you're correct. Invert the variables, though. It should be `if __name__ == "__main__"` .

ParthGandhi:
Was asked to review it again. Issues that I still see:
1. Still uses Python 3.x (each file begins with "#! /usr/bin/env python3").
2. Doesn't follow all pep-8 guidelines. For example, method names should use "_" between words and not be camel-case. 
3. Abstract classes don't use ABCs. If you don't intend to do this, please give a reason explaining why.

kpovetkin:
1. http://legacy.python.org/dev/peps/pep-0008/#naming-conventions
   ...
   The following naming styles are commonly distinguished:
   ...
   mixedCase (differs from CapitalizedWords by initial lowercase character!)
   ...
2. There are no abstract classes.

kpovetkin:
The way source code was tested for pep8 issues:
pep8 --show-source --show-pep8 *.py

Please correct me if it's wrong.

kpovetkin:
Python 2.6+ support added

ParthGandhi:
Added some feedback/questions via Hipchat for ease. Will update here with results when we come to a consensus.

sreevatsanraman:
@ParthGandhi  can this be merged?

stream-client/stream-client-python/v2/serviceconnector.py
```diff
@@ -0,0 +1,84 @@
+#! /usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+import requests
+from config import Config
+
+
+class NoFoundError(Exception):
```
ParthGandhi: `NotFoundError`

stream-client/stream-client-python/v2/serviceconnector.py
```diff
@@ -0,0 +1,84 @@
...
+
+    __errorCode = -1
```
ParthGandhi: Shouldn't need to be declared outside the constructor.

stream-client/stream-client-python/v2/streamclient.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Return value:
+        Time-To-Live property in seconds
+        """
+        uri = self.__prepare_uri(u'info', data=stream)
+        response = self.check_response_errors(
+            self.__serviceConnector.request(u'GET', uri)
+        )
+
+        ttl = response.json()[u'ttl']
```
ParthGandhi: Return here instead of creating a new local variable.

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+        }
+        """
+        dataToSend = None
+        headersToSend = None
+
+        if u'message' not in dataDict and u'file' not in dataDict:
+            raise TypeError(u'parameter should contain "message" \
+                            or "file" field')
+
+        if u'message' in dataDict and not isinstance(dataDict[u'message'], unicode) \
```
ParthGandhi: You can do `isinstance(dataDict['message'], (unicode, str))` to test against multiple types. Here and in the next couple of conditionals. 

https://stackoverflow.com/questions/11204789/how-to-properly-use-pythons-isinstance-to-check-if-a-variable-is-a-number

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+        u"""
+        Sets up handlers for successful and error responses.
+
+        Keyword arguments:
+        ok -- Handler to be called in case of successful response.
+        error -- Handler to be called in case of failed response.
+                 Could be of type None.  In that case would be the _same_
+                 as for a successful case.
+
+        Handlers should be a function daclared with next signature:
```
ParthGandhi: "Handlers should be a function declared with next signature:"

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+            finally:
+                self.__onOkHandler = self.__onErrorHandler = None
+        self.__handlersLock.release()
+
+    def on_response(self, ok, error=None):
+        u"""
+        Sets up handlers for successful and error responses.
+
+        Keyword arguments:
+        ok -- Handler to be called in case of successful response.
```
ParthGandhi: Rename to `success_handler` or similar.

stream-client/stream-client-python/v2/streamwriter.py
```diff
@@ -0,0 +1,72 @@
+#! /usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+import locale
+from serviceconnector import ServiceConnector, ConnectionErrorChecker
+from streampromise import StreamPromise
+
+
+class StreamWriter(object):
+    __serviceConnector = None
```
ParthGandhi: Again, don't think you need these declared outside, do you?

stream-client/stream-client-python/v2/streamwriter.py
```diff
@@ -0,0 +1,72 @@
...
+
+        Keyword arguments:
+        file -- path to file to be sent to Gateway
+        mimetype -- mimetype of a file. If is not defined would performed a try
+                    to detect mimetype automaticaly.
+
+        Returns:
+        StreamPromise instance for further handling
+        """
+        dataForPromise = {
```
ParthGandhi: Rename to `promiseData`.

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+        serviceConnector -- reference to connection pool to communicate with
+                            gateway
+        uri -- REST URL part to perform request.
+               Example: '/v2/strems/mystream'
+        data -- data to proceed by worker thread.
+                Please read '__workerTarget' documentation.
+        """
+        if not isinstance(serviceConnector, ServiceConnector):
+            raise TypeError(u'"serviceConnector" parameter \
+                            should be of type ServiceConnector')
```
ParthGandhi, line 29: This is short enough to be on the same line.

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+                                 determine mimetype itself.
+            'headers': dict      Additional HTTP headers.
+        }
+        """
+        dataToSend = None
+        headersToSend = None
+
+        if u'message' not in dataDict and u'file' not in dataDict:
+            raise TypeError(u'parameter should contain "message" \
+                            or "file" field')
```
ParthGandhi, line 72: Same line here too.

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+
+        if u'message' in dataDict and not isinstance(dataDict[u'message'], unicode) \
+           and not isinstance(dataDict[u'message'], str):
+            raise TypeError(u'"message" field should be of type \
+                            "string" or "bytes"')
+
+        if u'message' in dataDict and isinstance(dataDict[u'message'], unicode) \
+           and u'charset' not in dataDict:
+            raise TypeError(u'parameter should contain "charset" field \
```
ParthGandhi, line 80: And here.

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+                            in case if "message" is string')
+
+        if u'headers' in dataDict and dataDict[u'headers'] \
+           and not isinstance(dataDict[u'headers'], dict):
+            raise TypeError(u'"headers" field should be of type dict')
+
+        u"""
+        In case if message is of type 'bytes' it would not be rewriten
```
ParthGandhi: "rewritten"

ParthGandhi:
Just a few more minor formatting and doc comments. I've commented only on the version 2 implementation, but the same comments must be addressed in the version 3 implementation as well.

checkstyle.xml
```diff
@@ -0,0 +1,392 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright 2014 Cask, Inc.
```
awholegunch, line 3: Cask Data, Inc. (common)

pom.xml
```diff
@@ -0,0 +1,207 @@
...
+  ~ the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>co.cask.cdap</groupId>
+  <artifactId>cdap-injest</artifactId>
```
awholegunch, line 24: cdap-ingest [common]

pom.xml
```diff
@@ -0,0 +1,207 @@
...
+  <version>1.0-SNAPSHOT</version>
+  <packaging>pom</packaging>
+  <name>CDAP injest</name>
```
awholegunch, line 27: <name>CDAP Stream Ingestion</name>

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+- create a Stream with a specified <stream-id>;
+- update TTL for an existing Stream with a specified <stream-id>;
+- retrieve the current Stream TTL for a specified <stream-id>;
+- truncate an existing Stream (the deletion of all events that were written to the Stream);
+- write an event to an existing Stream specified by <stream-id>;
+- send a File to an existing Stream specififed by <stream-id>.
+
+## Usage
+
+ To use the Stream Client Python API, include this in your Python script:
```
awholegunch: To use the Stream Client Python API, include these imports in your Python script:

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+    from streamclient import StreamClient
+```
+
+## Example
+
+Create a StreamClient instance, specifying the fields 'host' and 'port' of the gateway server. 
+Optional configurations that can be set (and their default values):
+
+  - SSL: False (use HTTP protocol)
+  - apiKey:  '' (Need to specify to authenticate client requests using SSL)
```
awholegunch: - apiKey:  '' (need to specify only for authenticating ...

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+   ttl = streamClient.get_ttl("streamName");
+ ```
+
+ Create a ```StreamWriter``` instance for writing events to the Stream "streamName":
+
+ ```
+   streamWriter = streamClient.create_writer("streamName");
+ ```
+
+ To write new events to the Stream, you can use any of these two methods of the ```StreamWriter``` class:
```
awholegunch: To write new events to the Stream, use either of these methods of the `StreamWriter` class:

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+ To truncate the Stream *streamName*, use:
+
+ ```
+   streamClient.truncate("streamName");
+ ```
+
+ ### StreamPromise
+ StreamPromise's goal is to implement deferred code execution.
+
+For error handling, create a handler for each case and set it using the onResponse method.
```
awholegunch: and set it using the `onResponse` method.

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+   streamClient.create("newStreamName");
+ ```
+
+ Notes:
+
+  - The <stream-id> should only contain ASCII letters, digits and hyphens.
+  - If the Stream already exists, no error is returned, and the existing Stream remains in place.
+
+
+ Update TTL for the Stream "streamName"; TTL is a long value:
```
awholegunch: Update the TTL for the Stream "streamName", where `newTTL` is a long value:

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+ Create a new Stream with the *stream-id* "newStreamName":
+
+ ```
+   streamClient.create("newStreamName");
+ ```
+
+ Notes:
+
+  - The <stream-id> should only contain ASCII letters, digits and hyphens.
```
awholegunch: The <stream-id> can only contain ASCII letters, digits and hyphens.

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+ ```
+   config = Config()
+   config.host = 'localhost'
+   config.port = 10000
+   config.ssl = False
+
+   streamClient = StreamClient(config)
+ ```
+
+ or using the readFromFile method of Config object:
```
awholegunch: or using the `readFromFile` method of the `Config` object:

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+ To use the Stream Client Python API, include this in your Python script:
+
+```
+    from config import Config
+    from streamclient import StreamClient
+```
+
+## Example
+
+Create a StreamClient instance, specifying the fields 'host' and 'port' of the gateway server.
```
awholegunch: Create a `StreamClient` instance

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
+# CDAP Ingest Python library
+
+Stream Client Python API for managing streams via external Python applications.
```
awholegunch: managing Streams via

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
+# CDAP Ingest Python library
```
awholegunch: # CDAP Stream Client Python library

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+
+        self.__workerThread = Thread(target=self.__worker_target,
+                                     args=(uri, data))
+        self.__handlerThread = Thread(target=self.__response_check_target)
+        self.__workerThread.start()
+        self.__handlerThread.start()
+
+    def __worker_target(self, uri, dataDict):
+        u"""
+        Represents logic for performing requests and repsonses handling.
```
awholegunch: Represents logic for performing request and response handling.
This method should be invoked in a separate thread to reduce main
thread locks.

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+        This method should be invoked in separate thread to reduce main
+        thread locks.
+
+        uri -- REST URL part to perform request.
+               Example: '/v2/strems/myStream'
+        dataDict -- parameters that should be passed to REST server:
```
awholegunch: dataDict -- parameters that are to be passed to the REST server:

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+                )
+            except NoFoundError:
+                self.__onErrorHandler(self.__serviceResponse)
+            finally:
+                self.__onOkHandler = self.__onErrorHandler = None
+        self.__handlersLock.release()
+
+    def on_response(self, ok, error=None):
+        u"""
+        Sets up handlers for successful and error responses.
```
awholegunch: Sets up handlers for success and error responses.

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Keyword arguments:
+        ok -- Handler to be called in case of successful response.
+        error -- Handler to be called in case of failed response.
+                 Could be of type None.  In that case would be the _same_
```
awholegunch: In that case would be identical to a successful response.

stream-client/stream-client-python/v2/test/runtests.py
```diff
@@ -0,0 +1,284 @@
+#! /usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# Python 2.6 does not have 'unittest' module.
```
awholegunch: Python 2.6 does not have 'unittest' module; try importing 'unittest2' instead.

stream-client/stream-client-python/v2/test/runtests.py
```diff
@@ -0,0 +1,284 @@
...
+# 'unittest2' should de imported instead.
+try:
```
awholegunch, line 6: Should this not be the other way around? (as unittest is present only in 2.7, and is in the standard dist, and unittest2 in 2.4 - 2.6, but only if added):
try: 
  import unittest as unittest
except ImportError:
    import unittest2 as unittest

stream-client/stream-client-python/v2/streamwriter.py
```diff
@@ -0,0 +1,72 @@
...
+        return StreamPromise(self.__serviceConnector, self.__serviceUri,
+                             dataForPromise)
+
+    def send(self, file, mimetype=None):
+        u"""
+        Sends the content of a file as multiple stream events.
+
+        Keyword arguments:
+        file -- path to file to be sent to Gateway
+        mimetype -- mimetype of a file. If is not defined would performed a try
```
awholegunch: If is not defined, will attempt to detect mimetype automatically.

stream-client/stream-client-python/v2/streamwriter.py
```diff
@@ -0,0 +1,72 @@
...
+    def write(self, message, charset=None, headers=None):
+        u"""
+        Ingest a stream event with a string as body.
+
+        Keyword arguments:
+        message -- Data to transmit to REST server.
+                   Could be of type None if file field is presented.
+        charset -- Message field content charset. Could be of type None.
+                   Default value: 'utf-8'
+        headers -- Additional HTTP headers. Should be of type 'dict'
```
awholegunch: Additional HTTP headers. Should be of type 'dict'.

stream-client/stream-client-python/v2/streamwriter.py
```diff
@@ -0,0 +1,72 @@
...
+class StreamWriter(object):
+    __serviceConnector = None
+    __serviceUri = None
+
+    def __init__(self, serviceConnector, uri):
+        u"""
+        Object constructor
+
+        Keyword arguments:
+        serviceConnector -- reference to connection pool to communicate
```
awholegunch, line 13: reference to connection pool to communicate with gateway server.

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,188 @@
...
+                                 is presented.
+            'charset': '',       Message field content charset.
+                                 Could be of type None.
+                                 Default value: 'utf-8'
+            'file': '',          Path to a file which should be trasmited
+                                 to REST server.
+                                 Could be of type None if message field
+                                 is presented.
+            'mimetype': '',      Mimetype of a file which should be transmited.
+                                 If is omitted, application would try to
```
awholegunch: If omitted,

awholegunch:
From https://google-styleguide.googlecode.com/svn/trunk/pyguide.html

> Most .py files do not need to start with a #! line. Start the main file of a program with #!/usr/bin/python with an optional single digit 2 or 3 suffix per PEP-394.
> This line is used by the kernel to find the Python interpreter, but is ignored by Python when importing modules. It is only necessary on a file that will be executed directly.

So, on those python file in the ingest: we can remove most of the starting shebangs (#! /usr/bin/env python2 +# -_\- coding: utf-8 -_-), as most of these files are never to be called from the command line. 

Just leave them for the few (two?) that are called directly, the test examples. Please use "#!/usr/bin/env" as that is the most portable form. Not all installs have python in /bin.

#!/usr/bin/env python2
# -_\- coding: utf-8 -_-
or
#!/usr/bin/env python3
# -_\- coding: utf-8 -_-

stream-client/stream-client-python/v2/streampromise.py
kpovetkin, line 29: pep8 tool says this line is too long( > 79 symbols )

stream-client/stream-client-python/v2/streampromise.py
kpovetkin, line 72: pep8 tool says this line is too long( > 79 symbols )

stream-client/stream-client-python/v2/streampromise.py
kpovetkin, line 80: pep8 tool says this line is too long( > 79 symbols )

checkstyle.xml
kpovetkin, line 3: This file is not a part of the pull request.

pom.xml
kpovetkin, line 24: This file is not a part of the pull request.

pom.xml
kpovetkin, line 27: This file is not a part of the pull request.

sreevatsanraman:
@ParthGandhi @awholegunch  Have you guys done the review of latest changes. Can this be merged?

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+    from config import Config
+    from streamclient import StreamClient
+```
+
+## Example
+
+Create a ```StreamClient``` instance, specifying the fields 'host' and 'port' of the gateway server. 
+Optional configurations that can be set (and their default values):
+
+  - SSL: False (use HTTP protocol)
```
awholegunch: given that the fields are different than the names, perhaps we should be more explicit. Is it `config.apiKey`?:
- ssl: Use SSL or HTTP protocol (default: False, use HTTP protocol)
- apiKey: '' (required only when authenticating client requests using SSL)

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,137 @@
...
+  - apiKey:  '' (need to specify only for authenticating client requests using SSL)
```
awholegunch: I don't see this key being used anywhere...is it?

stream-client/stream-client-python/v2/streampromise.py
ParthGandhi, line 29: Nvm, you're right. I'm used to the 120 characters we use for Java.

prinam:
@ashvayka @kpovetkin just a few more comments from John and Parth to be addressed before this can be merged.

kpovetkin:
Changes are done.

stream-client/stream-client-python/README.md
```diff
@@ -0,0 +1,136 @@
...
+   config.port = 10000
+   config.ssl = False
+
+   streamClient = StreamClient(config)
+ ```
+
+ or using the ```read_from_file``` method of the ```Config``` object:
+
+ ```
+   config = Config.read_from_file('/path/to/config.ini')
```
ParthGandhi: "/path/to/config.json"

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,178 @@
...
+        success_handler -- Handler to be called in case of successful response.
+        error_handler -- Handler to be called in case of failed response.
+                 Could be of type None.  In that case would be _identical_
+                 to a successful response.
+
+        Handlers should be a function declared with next signature:
+
+        def coolErrorHandler( httpResponseObject):
+            ...
+            fooling around with response
```
ParthGandhi: "Error handling response".

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __init__(self, serviceConnector, uri, data):
+        u"""
+        Object constructor
+
+        Keyword arguments:
+        serviceConnector -- reference to connection pool to communicate with
+                            gateway
+        uri -- REST URL part to perform request.
+               Example: '/v2/strems/mystream'
```
ParthGandhi: "/v2/streams/mystream".

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,178 @@
...
+        self.__handlerThread.start()
+
+    def __worker_target(self, uri, dataDict):
+        u"""
+        Represents logic for performing requests and repsonse handling.
+        This method should be invoked in a separate thread to reduce main
+        thread locks.
+
+        uri -- REST URL part to perform request.
+               Example: '/v2/strems/myStream'
```
ParthGandhi: "/v2/streams/mystream"

stream-client/stream-client-python/v2/streampromise.py
```diff
@@ -0,0 +1,178 @@
...
+
+        self.__workerThread = Thread(target=self.__worker_target,
+                                     args=(uri, data))
+        self.__handlerThread = Thread(target=self.__response_check_target)
+        self.__workerThread.start()
+        self.__handlerThread.start()
+
+    def __worker_target(self, uri, dataDict):
+        u"""
+        Represents logic for performing requests and repsonse handling.
```
ParthGandhi: "response"

ParthGandhi:
Couple of minor doc fixes. Please fix in both versions. 

Otherwise, LGTM.

kpovetkin:
done

kpovetkin:
Done 
----- Original Message -----

sreevatsanraman:
LGTM :+1:

PR #8, opened 2014-09-03
CYB-2: [NEW] [Stream Ingestion] Design and Implement Ruby API for ingesting into stream

ashvayka:
New PR that contains ruby changes
I suspect we will have merge conflicts if java version of stream client API will be merged first. Due to same pom.xml and other files. In order to make correct PR we need to accept at least CYB-15 (Java API) and make this PR based on that code.


sreevatsanraman:
PR - 9 supercedes. So closing this.

PR #12, opened 2014-09-09
[cyb-3] Promise class for task implementation

kpovetkin:


stream-client/stream-client-javascript/src/promise.js
```diff
@@ -0,0 +1,143 @@
...
+             * @param {any} value
+             */
+            notifyImpl = function (value) {
+                notify_value_stack.push(value);
+                fireNotify();
+            };
+
+        return {
+            then: thenImpl,
+            catch: catchImpl,
```
gaarf, line 137: `catch` is a reserved word, for this to work in some browsers it will have to be set as a string...

`this["catch"] = ...`

gaarf:
A test would be nice... https://github.com/promises-aplus/promises-tests

kpovetkin:
Test would be latter with test for other parts of the library.

kpovetkin:
I'm sorry, this PR have to be closed. It's created with wrong target branch.

PR #13, opened 2014-09-09
[cyb-4] License information added

kpovetkin:
License information added to sources.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -240,6 +240,11 @@
                   <exclude>stream-client-ruby/spec/fixtures/1MB</exclude>
                   <exclude>stream-client-ruby/Gemfile.lock</exclude>
                   <exclude>stream-client-ruby/README.md</exclude>
+                  <exclude>stream-client-python/v2/test/*</exclude>
+                  <exclude>stream-client-python/v2/config.json</exclude>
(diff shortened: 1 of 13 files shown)


stream-client/stream-client-python/v2/config.py
```diff
@@ -1,3 +1,18 @@
+#  Copyright 2014 Cask Data, Inc.
```
awholegunch: Copyright © 2014 Cask Data, Inc. [common]

pom.xml
```diff
@@ -240,6 +240,11 @@
                   <exclude>stream-client-ruby/spec/fixtures/1MB</exclude>
                   <exclude>stream-client-ruby/Gemfile.lock</exclude>
                   <exclude>stream-client-ruby/README.md</exclude>
+                  <exclude>stream-client-python/v2/test/some.log</exclude>
```
awholegunch: You can use wildcards here to make it more robust, if that's your intent:
<exclude>stream-client-python/v2/test/*.log</exclude>

awholegunch:
A couple of comments, once addressed will be LGTM.

kpovetkin:
Done

awholegunch:
LGTM

PR #16, opened 2014-09-12
Cyb 74 update deb and rpm builds

ghost:
Update deb and rpm builds. See the last commit "CYB-74: FileTailer change deb/rpm"


file-tailer/src/main/deb/control/control
```diff
@@ -16,9 +16,11 @@
...
 Section: misc
+License: Copyright ${package.build.year} Cask Data, Inc. Licensed under the Apache License, Version 2.0.
 Priority: optional
 Architecture: all
+Url: http://cask.co
 Depends:
-Maintainer: cask
-Description: File Tailer
+Maintainer: support@cask.co
+Description: Component of Cask DAP
```
sreevatsanraman, line 25: Please use CDAP consistently in all places.

sreevatsanraman:
I don't see rpm and debian changes in this PR. We would want to make the rpm and debian builds consistent with CDAP. I don't see that here. What is this PR supposed to contain? :-1:

sreevatsanraman:
I also looked at the develop branch under file-tailer/pom.xml that doesn't have the change either.

sreevatsanraman:
I will merge this one. But we need to make the rpm and debian building consistent with CDAP. Will open a JIRA.

PR #23, opened 2014-09-16
Cyb 84 Python authentication library integrated into Python CDAP library. Integration tests added.

kpovetkin:
Python authentication library integrated into Python CDAP library.
Integration tests added.


file-tailer/src/main/java/co/cask/cdap/filetailer/config/PipeConfigurationImpl.java
```diff
@@ -26,6 +26,7 @@
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.lang.reflect.InvocationTargetException;
```
ParthGandhi, line 29: Java changes shouldn't be in this PR.

file-tailer/src/test/java/co/cask/cdap/filetailer/state/FileTailerStateProcessorImplTest.java
```diff
@@ -21,7 +21,6 @@
 import org.junit.Test;
```
ParthGandhi, line 21: Java changes.

stream-client/python/streamclient/v2/config.py
```diff
@@ -12,18 +12,23 @@
 #  License for the specific language governing permissions and limitations under
 #  the License.
 
-
 from __future__ import with_statement
 import json
 from io import open
+from caskauthclient.BasicAuthenticationClient import BasicAuthenticationClient
```
ParthGandhi, line 20: Hm, confused about where this is coming from.. I don't see this class in the python src code. 

The only `BasicAuthenticationClient` I know of is in our Java code.

stream-client/python/streamclient/v2/config.py
```diff
@@ -12,18 +12,23 @@
...
 
 
 class Config(object):
 
+
```
ParthGandhi: Extra newline.

stream-client/python/streamclient/v2/config.py
```diff
@@ -49,15 +54,30 @@ def ssl(self):
...
         newConfig.host = jsonConfig[u'hostname']
         newConfig.port = jsonConfig[u'port']
         newConfig.ssl = jsonConfig[u'SSL']
 
+        authClient = BasicAuthenticationClient()
+        authClient.set_connection_info(newConfig.host, newConfig.port,
+                                              newConfig.ssl)
+        authClient.configure(filename)
+
+        newConfig.set_auth_client(authClient)
```
ParthGandhi: Add an additional constructor to this class that takes in all the necessary properties (i.e. the non-default ones). Then you can call that constructor instead of setting each of host, port, etc individually.

stream-client/python/streamclient/v2/config.json
```diff
@@ -0,0 +1,7 @@
+{
```
ParthGandhi: Rename the file to "default-config.json".

stream-client/python/streamclient/v2/config.py
ParthGandhi, line 20: Oh, I see you have another PR open against "cdap-clients" which implements this. Is that where this is coming from?

stream-client/python/streamclient/v2/config.py
```diff
@@ -49,15 +54,30 @@ def ssl(self):
     def ssl(self, ssl):
         self.__ssl = ssl
 
+    @property
+    def auth_token(self):
+        try:
+            return self.__authClient.get_access_token()
```
ParthGandhi, line 75: Not sure how the auth_token is gotten in the other PR, but you may want to use a cached value here that can be overriden using a boolean flag to re-get the token.

stream-client/python/streamclient/v2/test/runtests_reactor.py
```diff
@@ -0,0 +1,166 @@
...
+from config import Config
+from serviceconnector import NotFoundError
+from streamwriter import StreamWriter
+from streamclient import StreamClient
+from caskauthclient.BasicAuthenticationClient import BasicAuthenticationClient
+
+
+class TestStreamClient(unittest.TestCase):
+
+    __host = u'localhost'
```
ParthGandhi: What cluster are these tests run against? Do I need to have a singlenode instance running to run them?

stream-client/python/streamclient/v2/test/runtests_reactor.py
```diff
@@ -0,0 +1,166 @@
...
+    __port = 10000
+    __ssl = False
+    __BASE_URL = u'http://{0}:{1}/v2'.format(__host, __port)
+    __REQUEST_PLACEHOLDERS = {
+        u'streamid': u'<streamid>'
+    }
+    __REQUESTS = {u'streams': __BASE_URL + u'/streams'}
```
ParthGandhi: Key is more appropriately "base_stream_path"

stream-client/python/streamclient/v2/test/runtests_reactor.py
```diff
@@ -0,0 +1,166 @@
...
+
+    exit_code = 404
+
+    def setUp(self):
+        authClient = BasicAuthenticationClient()
+        authClient.set_connection_info(self.__host, self.__port, self.__ssl)
+        authClient.configure(u'config.json')
+
+        config = Config()
+        config.host = self.__host
```
ParthGandhi: Use the above mentioned constructor here.

stream-client/python/streamclient/v2/test/runtests_reactor.py
```diff
@@ -0,0 +1,166 @@
...
+        config.port = self.__port
+        config.ssl = self.__ssl
+        config.set_auth_client(authClient)
+
+        self.sc = StreamClient(config)
+
+    def test_reactor_successful_connection(self):
+        try:
+            self.sc.create(self.validStream)
```
ParthGandhi: No assert here? Is there a way to assert on the return status?

stream-client/python/streamclient/v2/test/runtests_reactor.py
```diff
@@ -0,0 +1,166 @@
...
+        self.assertRaises(
+            NotFoundError,
+            self.sc.set_ttl,
+            self.invalidStream,
+            ttl
+        )
+
+    def test_get_ttl_valid_stream(self):
+        try:
+            self.sc.get_ttl(self.validStream)
```
ParthGandhi: Assert on the returned value here.

stream-client/python/streamclient/v2/test/runtests_reactor.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self.assertEqual(response.status_code, 200)
+
+    def test_set_ttl_valid_stream(self):
+        ttl = 88888
+
+        try:
+            self.sc.set_ttl(self.validStream, ttl)
+        except NotFoundError:
+            self.fail(u'StreamClient.setTTL() failed')
```
ParthGandhi: Method should be "set_ttl()" in the error message.

stream-client/python/streamclient/v2/test/runtests_reactor.py
```diff
@@ -0,0 +1,166 @@
...
+            self.sc.set_ttl,
+            self.invalidStream,
+            ttl
+        )
+
+    def test_get_ttl_valid_stream(self):
+        try:
+            self.sc.get_ttl(self.validStream)
+        except NotFoundError:
+            self.fail(u'StreamClient.getTTL() failed')
```
ParthGandhi: Method should be "get_ttl()" in the error message.

stream-client/python/streamclient/v2/test/runtests_reactor_ssl.py
```diff
@@ -0,0 +1,166 @@
+#! /usr/bin/env python2
```
ParthGandhi, line 1: This class and the previous one are pretty much exactly the same except 1 of them uses SSL and the other doesn't. 

The better thing to do would be to have an abstract class that both of these extends, that has the common functionality and tests. Then these two classes should simply override the appropriate values (for example, one of the sets SSL to true and the other defaults to false) and run the common tests.

ParthGandhi:
Done with an initial review. Please fix both versions.

stream-client/python/streamclient/v2/config.py
kpovetkin, line 20: Yes, you are right.

stream-client/python/streamclient/v2/config.py
kpovetkin, line 75: As long we have authentication client it would be better to implement caching in it.

stream-client/python/streamclient/v2/test/runtests_reactor.py
kpovetkin: No, there is no. create method should not return any value nor throw an exception. The only exception there could be - underlying http errors.

stream-client/python/streamclient/v2/test/runtests_reactor.py
kpovetkin: The reason it's done in this way is that we cannot be sure what value would be returned by a server. Can we be sure the default TTL value would not be changed? Python 'unittest' module does not guarantee that would be execute in the order as their are added.

Catching an exception guarantees we can detect without side effects is method working correct or not.

stream-client/python/streamclient/v2/test/runtests_reactor.py
ParthGandhi: Ah, these are integration tests? So they _must_ be run against a cluster?

stream-client/python/streamclient/v2/test/runtests_reactor.py
ParthGandhi: Got it. Ignore that comment then.

stream-client/python/streamclient/v2/test/runtests_reactor.py
ParthGandhi: Actually, there are ways to get unittest to run in the order you'd like. 

Having said that, I think doing that may be overkill here. So what you're doing is OK.

kpovetkin: Thank you for an updte. I'll check how it could be done! 

----- Original Message -----

stream-client/python/streamclient/README.md
```diff
@@ -4,12 +4,12 @@ Stream Client Python API for managing Streams via external Python applications.
 
 ## Supported Actions
 
-- create a Stream with a specified <stream-id>;
-- update TTL for an existing Stream with a specified <stream-id>;
-- retrieve the current Stream TTL for a specified <stream-id>;
+- create a Stream with a specified "*stream-id*";
```
awholegunch: \- create a Stream with a specified *stream-id*; [common]

stream-client/python/streamclient/v2/test/runtests_reactor_ssl.py
kpovetkin, line 1: Implemented. Please review.

stream-client/python/streamclient/v2/test/runtests_reactor.py
```diff
@@ -0,0 +1,25 @@
...
+    import unittest as unittest
+import requests
+
+from basicreactor import BasicReactor
+
+
+class TestStreamClient(unittest.TestCase, BasicReactor):
+
+    def setUp(self):
+        self.host = u'166.78.96.4'
```
ParthGandhi: These should be gotten from a test config file, where it defaults to localhost. Otherwise this test won't run if that cluster is down.

stream-client/python/streamclient/v2/test/runtests_reactor_ssl.py
```diff
@@ -0,0 +1,25 @@
...
+    import unittest as unittest
+import requests
+
+from basicreactor import BasicReactor
+
+
+class TestStreamClient(unittest.TestCase, BasicReactor):
+
+    def setUp(self):
+        self.host = u'198.61.160.4'
```
ParthGandhi: Same here. This should be configurable.

stream-client/python/streamclient/HOWTO_RUN_TESTS.md
```diff
@@ -0,0 +1,22 @@
+# HOWTO Run tests
+
+Tests are located in ```test``` directory of the source tree root.
+
+## Unittest
+In command shell type this:
+```> python runtests.py```
+
+## Integration tests
+### Change server connection information in ```runtests_reactor*.py```:
```
ParthGandhi: Instead of asking user to directly edit the file, as stated before, we should get this from a config.json file.

stream-client/python/streamclient/v2/test/runtests.py
```diff
@@ -0,0 +1,308 @@
...
+    os.path.abspath(inspect.getfile(inspect.currentframe())))
+parentdir = os.path.dirname(currentdir)
+sys.path.insert(0, parentdir)
+
+from config import Config
+from serviceconnector import NotFoundError
+from streamwriter import StreamWriter
+from streamclient import StreamClient
+
+with mock.patch('__main__.Config.is_auth_enabled',
```
ParthGandhi, line 42: This was previously reviewed, wasn't it? 

The only change is that now we're mocking the config to have auth_enabled to false. Shouldn't this come from config and not as a mock? So the only thing I think that you should need change is that the config object created in `setup` takes in false for auth_enabled. 

Let me know if i'm missing something and its doing something other than that.

stream-client/python/streamclient/v2/test/config.json
```diff
@@ -0,0 +1,8 @@
+{
+    "hostname": "198.61.160.4",
```
ParthGandhi: Default to localhost.

stream-client/python/streamclient/v2/test/basicreactor.py
```diff
@@ -0,0 +1,191 @@
...
+
+from config import Config
+from serviceconnector import NotFoundError
+from streamwriter import StreamWriter
+from streamclient import StreamClient
+from caskauthclient.BasicAuthenticationClient import BasicAuthenticationClient
+from caskauthclient.Config import Config as AuthConfig
+
+# Should be used as parent class for integration tests.
+# In children __host, __port, __ssl properties have to be set.
```
ParthGandhi: Also add a comment on how `setUp` needs to be implemented in children classes and that these properties should be set there. That method should also call the `set_up` method defined in here.

stream-client/python/streamclient/v2/test/basicreactor.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @property
+    def ssl_cert_check(self):
+        return self.__ssl_cert_check
+
+    @ssl_cert_check.setter
+    def ssl_cert_check(self, state):
+        self.__ssl_cert_check = state
+
+    def set_up(self):
```
ParthGandhi: rename to `base_set_up`.

ParthGandhi:
A few more comments. Please address in both versions. 

Also, I think there are 2 things we should change in existing code:
1. When an `error_handler` is not provided, we default it to be the same as the `response_handler`. This doesn't make much sense. Instead, if either of those isn't provided, it should default to a no-op function. 
That is, if an error_handler is not provided, define it to be a default method that does nothing and simply passes. 
2. The `error_handler` and `response_handler` can only be set on a promise. This doesn't always make sense since a promise may finish executing before those values have been set on it. I think the smarter thing to do would be as: when `streamClient.create_writer("streamName")` is called, it should also optionally take in an `error_handler` and `response_handler` objects as well, like: `streamClient.create_writer("streamName", response_handler, error_handler)`. Then any promise that is created by this writer should use those 2 handlers by default.  Hopefully this makes sense. If not, let me know and we can discuss it further.

stream-client/python/streamclient/HOWTO_RUN_TESTS.md
```diff
@@ -0,0 +1,22 @@
+# HOWTO Run tests
```
ParthGandhi: Would prefer calling this file: `TESTS.md`.

stream-client/python/streamclient/v2/test/runtests.py
kpovetkin, line 42: 'is_auth_enabled' property wraps Authentication Libarary which performs checking if authentication is enabled on server side, so it can not be get from config file and should be mocked in unit tests.

stream-client/python/streamclient/HOWTO_RUN_TESTS.md
kpovetkin: 1. Fixed;
2. Promise does not rely on execution ending only. If event handlers are added later, they would be triggered anyway.

kpovetkin: PR has been updated. Please review. 
Changes are implemented for all comments. 

----- Original Message -----

kpovetkin:
1. Fixed.
2. Both handler would be triggered anyway. Doesn`t matter when they are setted up.

anwar6953:
Streampromise is currently not executing __onOkHandler, when __onErrorHandler is at default value, None.

Also, ServiceConnector's **init** is still using config.json as a default value, and so fails when that file is not there (it currently doesn't exist in this branch).

kpovetkin:
Corrected.

anwar6953:
Alright, agreed. The bug in on_promise does appear to be resolved. Thanks.

ParthGandhi:
:+1: 

@poornachandra @anwar6953 Merging as discussed offline.

PR #26, opened 2014-09-17
CYB-81 	  Implement authentication for CDAP Flume Sink, and fix start status bug in start sink method 

ytolochkevych:


cdap-flume/README.md
```diff
@@ -36,8 +36,37 @@ The CDAP Sink is a Flume Sink implementation using RESTStreamWriter to write eve
  CDAP Gateway server version:
  ```
  a1.sinks.sink1.version = v2
+
+ ```
+
+## Authentication Client
+
+ If you want to use authentication,add authentication client configuration parameters to sink configuration:
```
anew: blank missing after comma. 

Also, it should be "to the sink configuration".

cdap-flume/README.md
```diff
@@ -36,8 +36,37 @@ The CDAP Sink is a Flume Sink implementation using RESTStreamWriter to write eve
...
+ 
+ Classpath of authentication client class:
```
anew: It is not the classpath, it is the fully qualified class name

cdap-flume/README.md
```diff
@@ -36,8 +36,37 @@ The CDAP Sink is a Flume Sink implementation using RESTStreamWriter to write eve
...
+ ````
+
+## Authentication Client Example Configuration
+ 
  ```
  
+     # User name
+     security.auth.client.username=admin
+     # User password
+     security.auth.client.password=realtime
```
anew: please don't use realtime.

cdap-flume/README.md
```diff
@@ -36,8 +36,37 @@ The CDAP Sink is a Flume Sink implementation using RESTStreamWriter to write eve
...
+     # Host name that is used by authentication client
+     security.auth.client.gateway.hostname=localhost
+     # Host port name that is used by authentication client
```
anew: Is it just port? And how about using security.auth.server.hostname and security.auth.server.port? Because this specifies where the auth server runs, not where the client runs.

cdap-flume/README.md
```diff
@@ -36,8 +36,37 @@ The CDAP Sink is a Flume Sink implementation using RESTStreamWriter to write eve
...
+ Path to authentication client properties file:
+ ````
+ a1.sinks.sink1.authClientProperties = /usr/local/apache-flume/conf/auth_client.conf
+ ````
+
+## Authentication Client Example Configuration
+ 
  ```
  
+     # User name
```
anew, line 59: Are these properties inherited from auth client? So you don't control them?

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -67,8 +92,21 @@ public void configure(Context context) {
     version = context.getString("version", DEFAULT_VERSION);
     writerPoolSize = context.getInteger("writerPoolSize", DEFAULT_WRITER_POOL_SIZE);
     streamName = context.getString("streamName");
+    authClientClassPath = context.getString("authClientClass", DEFAULT_AUTH_CLIENT);
```
anew: It is not a classpath but a class name.

anew: How would you disable authentication? If basic is the default, it won't work to specify nothing.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -83,7 +121,24 @@ public Status process() throws EventDeliveryException {
...
+            LOG.info("Success write to stream {}", streamName);
+          }
+
+          @Override
+          public void onFailure(Throwable t) {
+            writeStatus = EventWriteStatusCode.FAIL;
+            LOG.error("Error during writing event to stream {}", streamName, t);
+          }
+        });
+
```
anew: since you don't wait for the future to complete, execution in this thread will move on. So now you will check for FAIL but because the future is still executing, writeStatus will be whatever it was before (left over from previous invocation of process()?). Most likely it is OK and so you will commit the transaction and return, before the future is completed. That is wrong.

anew: Also, isn't the contract of flume that process() is synchronous? So you have to wait for the future to complete.

anew: And because writeStatus is a class member, all asynchronous writes will compete for that. If you want to make it async, each invocation must have its own private copy of writeStatus, and the transaction closing and other handling must move to the callback. Now I don't know what Flume will do if you return before the write is complete. It is probably safer to make this synchronous by calling get() on the future. I am happy to discuss this in person.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -128,7 +185,31 @@ private void createStreamClient() throws IOException {
 
       builder.writerPoolSize(writerPoolSize);
       builder.version(version);
+      AuthenticationClient authClient = null;
+      InputStream inStream = null;
+      try {
+        authClient = (AuthenticationClient) Class.forName(authClientClassPath).getConstructor().newInstance();
```
anew: getConstructor() is not needed here. (getConstructor is only useful if you need to pass in arguments)

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -128,7 +185,31 @@ private void createStreamClient() throws IOException {
...
+        authClient.setConnectionInfo(host, port, sslEnabled);
+        Properties properties = new Properties();
+        inStream = new FileInputStream(authClientPropertiesPath);
+        properties.load(inStream);
+        authClient.configure(properties);
+        builder.authClient(authClient);
+      } catch (ReflectiveOperationException e) {
+        LOG.error("Can not resolve class {}: {}", new Object[]{authClientClassPath, host, port, e.getMessage(), e});
```
anew: So you just log and ignore the error? That is, you will proceed as if no authentication was configured.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -170,17 +253,32 @@ public synchronized void start() {
...
     }
-    super.start();
     LOG.info("StreamSink {} started.", getName());
+    lifecycleState = LifecycleState.START;
   }
 
   public synchronized void stop() {
     LOG.info("StreamSink {} stopping...", getName());
     closeWriterQuietly();
     closeClientQuietly();
```
anew, line 254: does the auth client have to closed? Or does the streamClient do that when it closes?

anew:
Done with this review. There is some serious trouble in the new asynchronous stream client calls. Please also see my other comments. Thanks -Andreas.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
ytolochkevych: authentication is disabled If you do not specify path to auth client properties.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
ytolochkevych: yes, if authentication is disabled, sink will try to create writer without authentication

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
ytolochkevych, line 254: It is care of the streamClient

cdap-flume/README.md
ytolochkevych: changed

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
ytolochkevych: Made this synchronous

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
ytolochkevych: deleted

ytolochkevych:
I have updated pull request after review

cdap-flume/README.md
```diff
@@ -36,7 +36,36 @@ The CDAP Sink is a Flume Sink implementation using RESTStreamWriter to write eve
  CDAP Gateway server version:
  ```
  a1.sinks.sink1.version = v2
+
+ ```
+
+## Authentication Client
+
+ If you want to use authentication, add authentication client configuration parameters to sink the configuration:
```
awholegunch: To use authentication, add these authentication client configuration parameters to the sink configuration file:

ashvayka:
@awholegunch @sreevatsanraman I have merged develop branch and fixed documentation comments

cdap-flume/README.md
```diff
@@ -36,7 +36,36 @@ The CDAP Sink is a Flume Sink implementation using RESTStreamWriter to write eve
...
+
+     # User name
+     security.auth.client.username=admin
+     # User password
+     security.auth.client.password=secret
+     # Host name that is used by authentication client
+     security.auth.server.hostname =localhost
+     # Host port name that is used by authentication client
+     security.auth.server.port=10000
+     # SSL mode [true|false]
```
anew: I am wondering whether this should also be changed to 
security.auth.server.ssl=false
because it is also a server property, right?

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
anew: That seems a strange design to me. The primary control for the auth client is the actual auth client class. Disabling should be done by not specifying an auth client. The properties are a secondary parameter that is only needed if auth is on. Can we say, to disable auth, you specify empty string or a special string such as "none" for the auth client class?

anew: The way it is implemented now - correct me if I am wrong - would be: 
1. load the auth client class 
2. load the properties from file input stream
3. configure the client with the properties.
If no auth properties are specified, then the file name defaults to "", so the step 2. will throw an IOException, which you will then log as an error and proceed. So whenever I don't want auth to be turned on, I will see an Error in the logs. Not how it should be.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -83,7 +123,24 @@ public Status process() throws EventDeliveryException {
       transaction.begin();
       Event event = channel.take();
       if (event != null) {
-        writer.write(ByteBuffer.wrap(event.getBody()), event.getHeaders());
+        ListenableFuture future = writer.write(ByteBuffer.wrap(event.getBody()), event.getHeaders());
```
anew: What is the type parameter of the future returned by write()? 
I think the right way to do this is something like this:

``` java
try {
  writer.write(ByteBuffer.wrap(event.getBody()), event.getHeaders()).get();
} catch (Throwable e) (
  if (e instanceof ExecutionException) {
    e = e.getCause();
  }
  LOG.error("Error during writing event to stream {}", streamName, t);
  throw new EventDeliveryException("Failed to send events to stream: " + streamName, t);
}
transaction.commit();
...
```

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
anew: see my comment above. I think this is not the right way to do that. If no auth client was specified, we should not try to create one. If one was specified, we should create and configure it, and any failure means the StreamSink failed to start up.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
anew, line 254: Hmmm looking at the RestStreamClient, it does not seem to do that. Can you please verify?
Also, in general it is good practice to follow, that whoever opens a closeable is also responsible for closing it. In this, because the auth client is created and configured by this class, it should also be closed here. What do you think?

anew:
I have made further comments. Please address. Thanks.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
ytolochkevych, line 254: hm, It is bug in authentication client  implementation. It does not have have method, which close connection correctly. I have created new pull request  https://github.com/caskco/cdap-clients/pull/11 for this

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
ytolochkevych: ok

cdap-flume/README.md
ytolochkevych: security.auth.server.host,   security.auth.server.port, security.auth.server.ssl  are not used, because authentication send initial request to cdap server, and receive required auth_url in the answer. I deleted this parameters

anew: I see. That is much better.

anew:
Thanks for addressing the comments. I have just a few more comments that I missed in the last round. Please see the comments I made on the last commit.  Thanks.

ytolochkevych:
I have updated  pull request. Please review. Thanks

anew:
Looks good now. Thanks for the quick fix!

PR #34, opened 2014-09-22
Fix stream-client unit tests

jwang47:


stream-client/stream-client-java/src/test/java/co/cask/cdap/client/rest/RestClientTest.java
```diff
@@ -70,7 +64,7 @@ public void testOkResponseCodeAnalysis() {
     verify(response).getStatusLine();
   }
 
-  @Test(expected = BadRequestException.class)
+  @Test(expected = HttpFailureException.class)
```
poornachandra, line 67: To be more precise you may want to catch the exception and assert that the right status code is returned. Like -

```
    try {
      RestClient.responseCodeAnalysis(response);
      fail("Expected bad request status.");
    } catch (HttpFailureException e) {
      assertEquals(HttpURLConnection.HTTP_BAD_REQUEST, e.getStatusCode());
    }
```

poornachandra:
Minor comment, otherwise LGTM :+1:

sreevatsanraman:
LGTM :+1:

PR #35, opened 2014-09-22
Changed stream-client tests to use a better way to expect exception in a...

jwang47:
... test


stream-client/stream-client-java/src/test/java/co/cask/cdap/client/rest/RestStreamClientTest.java
```diff
@@ -51,22 +52,40 @@ public void testSuccessGetTTL() throws IOException {
     assertTrue(ttl == STREAM_TTL);
   }
 
-  @Test(expected = HttpFailureException.class)
+  @Test
   public void testNotFoundGetTTL() throws IOException {
-    streamClient.getTTL(TestUtils.NOT_FOUND_STREAM_NAME);
+    TestUtils.verifyException(HttpFailureException.class, new Callable<Void>() {
```
poornachandra: Sorry for being picky - but this too can take in a status code (404 in this case as it is NOT_FOUND) instead of HttpFailureException.class, right? This will make the tests more accurate.

poornachandra:
LGTM :+1:

PR #40, opened 2014-09-23
Feature/fix java example

anwar6953:
The java ingest example was throwing an exception because it was trying to create a StreamWriter for a stream that does not exist (it was creating the stream later).

I simply reordered the creation of the stream to before the creation of the StreamWriter for this stream.


stream-client/stream-client-java-example/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -32,7 +32,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(Main.class);
 
   public static void main(String[] args) {
-    final String streamName = "streamName";
+    final String STREAM_NAME = "exampleStream";
```
poornachandra: STREAM_NAME can be moved out as a static member variable since it is a constant.

poornachandra:
One minor comment, please fix and merge :+1:

PR #41, opened 2014-09-23
Feature/remove stream client send file

gsps1:
Removing send file method from stream client. 

sendFile to stream, sends the entire file as one event, doesn't implement what the doc says. so we are removing it for now.


ParthGandhi:
:+1:

gsps1:
@ParthGandhi can you take a quick look at this again?

anwar6953:
LGTM

PR #43, opened 2014-09-23
Added verifySSLCert functionality to cdap-flume

jwang47:


integration-tests/stream-client-java-integration-tests/src/test/java/co/cask/cdap/client/StreamClientIT.java
```diff
@@ -65,7 +65,7 @@ private StreamClient getTestClient() throws IOException {
     RestStreamClient.Builder clientBuilder = RestStreamClient.builder(properties.getProperty("host"),
                                                                       Integer.valueOf(properties.getProperty("port")));
     clientBuilder.ssl(Boolean.valueOf(properties.getProperty("ssl", "false")));
-    clientBuilder.disableCertCheck(Boolean.valueOf(properties.getProperty("disableCertCheck", "false")));
+    clientBuilder.verifySSLCert(Boolean.valueOf(properties.getProperty("disableCertCheck", "false")));
```
poornachandra, line 68: The default value should be "true"

jwang47, line 68: This is for the integration test, so this shouldn't be an issue.

poornachandra:
Minor comment, fix and merge :+1:

PR #44, opened 2014-09-24
Refactor

sreevatsanraman:
- Naming all artifacts consistently with cdap- prefix.


suppressions.xml
```diff
@@ -22,6 +22,8 @@
 
 <suppressions>
 
+  <suppress checks=".*" files="Gemfile.lock" />
```
poornachandra: This can be removed.

suppressions.xml
```diff
@@ -30,13 +32,5 @@
...
-  <suppress checks=".*" files=".*/LocalJobRunnerWithFix.java" />
-
-  <!-- do not check thrift generated files -->
-  <suppress checks=".*" files=".*/transaction/distributed/thrift/.*" />
-
-
-
-</suppressions>
\ No newline at end of file
+  <suppress checks=".*" files=".*/Gemfile.lock" />
```
poornachandra: This can be removed.

stream-client/stream-client-java-example/src/main/java/co/cask/cdap/client/example/Main.java
```diff
@@ -1,94 +0,0 @@
...
-import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-/**
- * An example of using the StreamClient API.
- */
-public class Main {
```
poornachandra, line 31: Is this file not required?

stream-client/stream-client-java-example/README.md
```diff
@@ -1,7 +0,0 @@
-Stream Client Java Example
```
poornachandra, line 1: Is this file not required?

suppressions.xml
sreevatsanraman: Fixed

suppressions.xml
sreevatsanraman: Fixed

stream-client/stream-client-java-example/src/main/java/co/cask/cdap/client/example/Main.java
sreevatsanraman, line 31: This is not adding much value. It is just one File that shows how to use. This should be in a README. Nooone uses it.

poornachandra:
Minor comments - LGTM :+1:

stream-client/stream-client-java-example/README.md
sreevatsanraman, line 1: Same as the other. This file is a README like file. Not porting.

sreevatsanraman:
Addressed comments merging.

PR #53, opened 2014-09-30
[cyb-3] JS ingest library implemented

kpovetkin:


stream-client/stream-client-javascript/Gruntfile.js
```diff
@@ -0,0 +1,74 @@
...
+        // Metadata
+        pkg: grunt.file.readJSON('package.json'),
+        // Task configuration
+        // Browser side tests
+        mocha: {
+            browser: {
+                src: ['test/client.html'],
+                options: {
+                    run: true,
+                    reporter: 'Nyan',
```
ajainarayanan: Lol.. Nyan reporter for mocha test :smiley:

gaarf:
comes with tests, love it! Will check out asap.

anwar6953:
This JS stream-ingest library is not yet integrated with the JS auth-client, correct?
(https://github.com/cybervisiontech/cdap-clients/commit/185350a115d77308c15ad68f53dd760e5717d0c9)
(So it does not currently work with secure clusters)

Is the integration a work in progress?

stream-client/stream-client-javascript/src/serviceconnector.js
```diff
@@ -0,0 +1,161 @@
...
+                        'Request not initialized',
+                        'Server connection established',
+                        'Request received',
+                        'Processing request',
+                        'Request finished and response is ready'
+                    ];
+
+                    promise.notify(readyStates[httpCon.readyState]);
+
+                    if (httpCon.readyState === 4) {
```
anwar6953: can you replace 4 with `XMLHttpRequest.DONE`

stream-client/stream-client-javascript/README.MD
```diff
@@ -0,0 +1,46 @@
+# CDAP Client JavaScript library
+
+Client JavaScript API for sending request to CASK reactor.
```
anwar6953: Just a reminder, we have renamed reactor to CDAP (which is an acronym for Cask Data Application Platform).

anwar6953:
Take a look at the current directory structure at:
https://github.com/caskdata/cdap-ingest

All these files should likely go under `cdap-stream-clients/javascript` instead of `stream-client/stream-client-javascript`

stream-client/stream-client-javascript/src/promise.js
```diff
@@ -0,0 +1,167 @@
...
+             * @param {function} handler
+             */
+            catchImpl = function (handler) {
+                return thenImpl.apply(this, [null, handler]);
+            },
+
+            /**
+             * Resolves the derived promise with the value.
+             *
+             * @param {any} value
```
anwar6953: Why put brackets around "any"? Shouldn't brackets be used only if its a type?

Same question for other 2 usages of `{any}`.

stream-client/stream-client-javascript/src/serviceconnector.js
```diff
@@ -0,0 +1,161 @@
...
+            baseUrl = function () {
+                return ['', (server.ssl ? 'https' : 'http'), '://',
+                    server.hostname, ':', server.port].join('');
+            },
+
+            /**
+             * @param {object} requestParams
+             * Possible fields for requestParams:
+             *
+             * @param {object} data         - data to send to a server
```
anwar6953: There doesn't appear to be a parameter in your function named `data`. It may be a field of requestParams, but that doesn't make it a @param. (Remove the annotation from data)

stream-client/stream-client-javascript/test/cdaptracker-browser.js
```diff
@@ -0,0 +1,102 @@
...
+            var tracker = CDAPTracker.ServiceConnector(
+                'localhost', 10000, false),
+            promise = tracker.track('/some/cool/url', {}),
+            promiseFired = false,
+
+            checkResult = function () {
+                expect(promiseFired).to.be.ok();
+            };
+
+            promise.then(null, null, function (data) {
```
anwar6953: can you remove data parameter?
If its not necessary, its better to remove it, right? (to avoid even minor confusion)

Please also apply this change to the three other instances where this is done.

stream-client/stream-client-javascript/test/cdaptracker-spec.js
```diff
@@ -0,0 +1,107 @@
...
+
+            promise.catch(function () {
+                promiseFired = true;
+            });
+
+            promise.then(checkResult, checkResult);
+        });
+
+        it('Checking promise notification handler( then(null, null, handler) )', function () {
+            var r = nock('http://localhost:10000')
```
anwar6953: Is there a need to assign it? Someone reading may look down for the variable `r`, only to find it unused. Better not to assign it, if unnecessary.

stream-client/stream-client-javascript/test/cdaptracker-spec.js
```diff
@@ -0,0 +1,107 @@
...
+
+            promise.then(function (data) {
+                promiseFired = true;
+            });
+
+            promise.then(checkResult, checkResult);
+        });
+
+        it('Checking promise "catch" method', function () {
+            var r = nock('http://localhost:10000')
```
anwar6953: Any reason for assignment to unused variable?

stream-client/stream-client-javascript/test/cdaptracker-spec.js
```diff
@@ -0,0 +1,107 @@
...
+
+            var promise = tracker.track('/some/cool/url', {});
+
+            expect(promise).to.be.an('object');
+            expect(promise).to.have.property('then');
+            expect(promise).to.have.property('catch');
+        });
+
+        it('Checking promise "then" method', function () {
+            var r = nock('http://localhost:10000')
```
anwar6953: Any reason for assignment to unused variable?

stream-client/stream-client-javascript/test/cdaptracker-spec.js
```diff
@@ -0,0 +1,107 @@
...
+            tracker = new CDAPTracker.ServiceConnector();
+            expect(tracker).to.have.property('track');
+        });
+    });
+
+    describe('Checking Tracker functionality', function () {
+        it('"track" method returns promise object', function () {
+            tracker = CDAPTracker.ServiceConnector();
+
+            var r = nock('http://localhost:10000')
```
anwar6953: Any reason for assignment to unused variable?

stream-client/stream-client-javascript/src/serviceconnector.js
```diff
@@ -0,0 +1,161 @@
...
+                request.end();
+
+                return promise;
+            },
+
+            request = ('undefined' != typeof window) ? requestBrowser : requestNode,
+
+            trackImpl = function (url, data) {
+                if (!(data instanceof Object)) {
+                    throw TypeError('"data" parameter has to be of type "Object"');
```
anwar6953: You should use new before the constructor.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError
http://js-bits.blogspot.com/2010/08/constructors-without-using-new.html

Please also apply this change to the other three places where you throw this error.

anwar6953:
This code seems to allow making requests from a browser. This is from the README:

## Usage

 To use the Client JavaScript API, include these scrip tag in your index.html:

In web browser:

```
<script type="text/javascript" src="cask-client.min.js"></script>
```

How will that circumvent the [Same Origin Policy](http://en.wikipedia.org/wiki/Same-origin_policy) to make requests?
For instance, if I execute the following in my browser's console when the browser is not at a google webpage, I would receive an error.
`f=new XMLHttpRequest(); f.open('GET', 'https://www.google.com/', false); f.send();`
Error: `XMLHttpRequest cannot load https://www.google.com/. No 'Access-Control-Allow-Origin' header is present on the requested resource.`

Wouldn't this library receive the same error when making requests?

stream-client/stream-client-javascript/README.MD
```diff
@@ -0,0 +1,46 @@
...
+
+## Supported Actions
+
+- sending request to gateway server
+
+## Usage
+
+ To use the Client JavaScript API, include these scrip tag in your index.html:
```
awholegunch: To use the Client JavaScript API, include these script tags in your index.html: [eliminates leading space, fixes spelling]

stream-client/stream-client-javascript/README.MD
```diff
@@ -0,0 +1,46 @@
...
+
+In web browser:
+```
+<script type="text/javascript" src="cask-client.min.js"></script>
+```
+
+In Node.JS:
```
awholegunch: In Node.js:

stream-client/stream-client-javascript/README.MD
awholegunch: Client JavaScript API for sending requests to the Cask Data Application Platform (CDAP).

stream-client/stream-client-javascript/README.MD
```diff
@@ -0,0 +1,46 @@
...
+
+## Tracker object
+Methods:
+
+  'track(url, data)' - makes POST request including *data* to *url*. *data* should be an object.
+                       Returns a promise object.
+
+## Promise object:
+
+  'then(successCallback, errorCallback, notifyCallback)     - regardless of when the promise was or will be resolved or rejected, then calls one of the success or error callbacks asynchronously as soon as the result is available. The callbacks are called with a single argument: the result or rejection reason. Additionally, the notify callback may be called zero or more times to provide a progress indication, before the promise is resolved or rejected.
```
awholegunch: Shouldn't this line be wrapped or broken?

stream-client/stream-client-javascript/TESTS.md
```diff
@@ -0,0 +1,13 @@
+#Tests
+
+## Browser
+
+Open 'test/SpecRunner.html' in your browser.
+
+## Node.JS
```
awholegunch: Node.js

stream-client/stream-client-javascript/README.MD
```diff
@@ -0,0 +1,46 @@
...
+
+  'catch(errorCallback)'        - shorthand for 'then(null, errorCallback)'
+  'finally(callback)'           – allows to observe either the fulfillment or rejection of a promise, but to do so without modifying the final value.
```
awholegunch: If the phrase does not begin with a capital, do not end with a period. Perhaps turn all these into sentences, as the first comment is long.

awholegunch: allows the observer either the fulfillment or rejection of a promise, but without modifying the final value
[is this what you mean?]

stream-client/stream-client-javascript/README.MD
awholegunch: Regardless of when the promise was (or will be) resolved or rejected, this calls one of either the success or error callbacks asynchronously as soon as the result is available. The callbacks are called with a single argument: the result or rejection reason. Additionally, the notify callback may be called zero or more times to provide a progress indication, before the promise is either resolved or rejected.

stream-client/stream-client-javascript/README.MD
```diff
@@ -0,0 +1,46 @@
...
+## Example
+
+Create a ```ServiceConnector``` instance, specifying the fields 'host' and 'port' of the gateway server. 
+Optional configurations that can be set (and their default values):
+
+  - ssl: False (use HTTP protocol)
+
+```
+    var tracker = new CDAPTracker.ServiceConnector('server_name', 10000),
+        pr = tracker.track('/path/to/request', {});
```
awholegunch: Indentation?

stream-client/stream-client-javascript/README.MD
```diff
@@ -0,0 +1,46 @@
...
+
+In Node.JS:
+```
+var CDAPTracker = require('serviceconnector');
+```
+
+## Tracker object
+Methods:
+
+  'track(url, data)' - makes POST request including *data* to *url*. *data* should be an object.
```
awholegunch: Makes POST request including _data_ to _url_. _data_ should be an object.

stream-client/stream-client-javascript/README.MD
```diff
@@ -0,0 +1,46 @@
+# CDAP Client JavaScript library
+
+Client JavaScript API for sending request to CASK reactor.
+
+## Supported Actions
+
+- sending request to gateway server
```
awholegunch: Send a request to the gateway server

gaarf:
@anwar6953 cross-origin is not within the scope of this PR. The lib's consumer will be responsible for supplying a proxy as necessary until the CDAP server has CORS support.

anwar6953:
Edit: see my note below: https://github.com/caskdata/cdap-ingest/pull/53#issuecomment-58400300

# Overall Usability Concern

My understanding is that the code in this pull request (PR) is _supposed to be_ a javascript implementation of [StreamClient](https://github.com/caskdata/cdap-ingest/tree/develop/cdap-stream-clients).

According to the README and my understanding of the code, the code in this PR will let you simply send POST requests to arbitrary endpoints in CDAP (**the user has to specify the urls and data**).

Shouldn't it implement an interface such as the [Java StreamClient's](https://github.com/caskdata/cdap-ingest/blob/develop/cdap-stream-clients/java/src/main/java/co/cask/cdap/client/StreamClient.java):

Users of this library shouldn't have to construct the url themselves for each of the following actions: creation of a stream, writing to a stream, truncation of a stream, or setting/getting the ttl of a stream. This set of functionality should be implemented as methods of the stream client, which the user can call with the appropriate parameters (stream name, message to ingest, ttl, etc). See link at bottom for implementation of these functions in java.

## Summary:

The way this library currently is, in order to take any actions on a CDAP stream, I have to construct the urls manually and send the request with the appropriate data. Another problem/limitation is that I can't get/set ttl, nor can I create streams, because GETs/PUTs are not allowed with what is exposed in this library.

How it should be: I should simply be able to call methods such as:

```
streamclient = new StreamClient('localhost', 10000);
streamclient.create("niceName"); // creates a stream in the locally running CDAP instance.
streamclient.truncate(“niceName”); //truncates a stream
 ... // other actions on a stream (set ttl, write to a string, etc)
```

[See the java streamclient and streamwriter interfaces and implementations, for example.](https://github.com/caskdata/cdap-ingest/tree/develop/cdap-stream-clients/java/src/main/java/co/cask/cdap/client)

anew:
I fully agree with Ali's concern.

anwar6953:
It seems that the discrepancy between the design/functionality in this PR as compared to the java version of the stream client is not unwarranted - there is a JIRA for the design of this to be this way.
https://jira.continuuity.com/browse/CYB-3

anwar6953:
Closing. Reopened here:
https://github.com/caskdata/cdap-ingest/pull/76

PR #54, opened 2014-10-01
cyb_95 integrate auth to ruby client

ytolochkevych:


cdap-stream-clients/ruby/lib/stream-client-ruby/rest.rb
```diff
@@ -38,15 +39,32 @@ def ssl?
...
 
+    def is_auth_enabled
+      self.class.auth_client.auth_enabled?
+    end
+
+    def get_access_token
+      self.class.auth_client.get_access_token
+    end
+
+    def self.set_auth_client auth_client
```
anwar6953: There should be a set_auth_client method on the StreamClient object as well (which might simply delegate to this), since the user should only be interacting with the streamclient, and not care about the rest object.

``` ruby
client = StreamClient.new
client.set_auth_client ...
```

instead of

``` ruby

client = StreamClient.new
client.rest.set_auth_client ...
```

(Try to make the interfaces as similar to java as possible).

ytolochkevych:
@anwar6953  agree. Changed

cdap-stream-clients/ruby/README.md
```diff
@@ -63,6 +63,11 @@ Create a StreamClient instance and use it as any Ruby object:
 client = CDAPIngest::StreamClient.new
 ```
 
+ if authentication is using, configure set authentication client:
+ ```
+  client.set_auth_client(auth_client)
```
anwar6953: Should this be done the more ruby way?
`client.set_auth_client auth_client`

ytolochkevych:
@anwar6953  I have updated PR after your comments

anwar6953:
@ytolochkevych
This looks good to me. :+1: 

This being merged should probably hold off though, until the ruby authentication client is merged in the cdap-clients repo, because of the dependency.

PR #56, opened 2014-10-03
CYB-79 Integration tests for CDAP flume sink 

ytolochkevych:


cdap-integration-tests/cdap-flume-java-integration-tests/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+      <groupId>org.apache.flume</groupId>
+      <artifactId>flume-ng-core</artifactId>
+      <version>1.5.0.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.flume</groupId>
+      <artifactId>flume-ng-embedded-agent</artifactId>
+      <version>1.5.0.1</version>
+    </dependency>
+    <dependency>
```
anew: you need two org.json:json? Why is one in test scope and another one for regular scope?

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+    for (int i = 0; i < eventNumber; i++) {
+      Event event = new SimpleEvent();
+      event.setBody((EVENT_STR + i).getBytes());
+      agent.put(event);
+    }
+    Thread.currentThread().sleep(1000);
+    agent.stop();
+    HttpResponse response = readFromStream(properties, startTime, System.currentTimeMillis());
+    JSONArray jsonArray = new JSONArray(EntityUtils.toString(response.getEntity()));
+    for (int i = 0; i < eventNumber; i++) {
```
anew: Assert.assertEquals(eventNumber, jsonArray.size()) would be good here.

cdap-integration-tests/cdap-flume-java-integration-tests/pom.xml
anew: We normally use Google's Gson for JSON de/serialization. It would be nice if you could switch.

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+import java.net.URI;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+
+public class CdapFlumeIT {
+
+  private static final int eventNumber = 50;
```
anwar6953: also capitalize this?

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+    field.setAccessible(true);
+
+    Field modifiersField = Field.class.getDeclaredField("modifiers");
+    modifiersField.setAccessible(true);
+    modifiersField.setInt(field, field.getModifiers() & ~Modifier.FINAL);
+    String sinkList[] = {"co.cask.cdap.flume.StreamSink"};
+    field.set(null, sinkList);
+  }
+
+  private HttpResponse readFromStream(Properties properties, long startTime, long endTime) throws Exception {
```
anew: is there a reason why you need to use apache.http for URL connections? Can't you just useJava's URL.openConnection()? We try to avoid pulling in extra libraries when it is not strictly required. 

If you look for more convenience, you could also use the HttpRequests in cdap-common.

cdap-integration-tests/pom.xml
```diff
@@ -35,6 +35,7 @@
     <module>cdap-stream-client-integration-tests</module>
     <module>cdap-file-tailer-integration-tests</module>
     <module>cdap-file-drop-zone-integration-tests</module>
+    <module>cdap-flume-java-integration-tests</module>
```
anew, line 38: why do we have a separate maven module for each integration test? It seems overkill to create a maven module for a single Java class. Can you refactor this? If it too much for this PR because it requires changing other tests, too, I am fine with a new PR for that.

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+    for (final String name : properties.stringPropertyNames()) {
+      propertyMap.put(name, properties.getProperty(name));
+    }
+    agent.configure(propertyMap);
+    agent.start();
+    long startTime = System.currentTimeMillis();
+    for (int i = 0; i < eventNumber; i++) {
+      Event event = new SimpleEvent();
+      event.setBody((EVENT_STR + i).getBytes());
+      agent.put(event);
```
anew: It's not clear to me how the stream gets created. And if you assume that the stream already exists, who creates it?

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+      Assert.assertTrue(jsonArray.getJSONObject(i).getString(RESPONSE_EVENT_BODY).equals(EVENT_STR + i));
+    }
+  }
+
+  @Test(expected = EventDeliveryException.class)
+  public void failEventProcessingTest() throws Exception {
+
+    EmbeddedAgent agent = new EmbeddedAgent("test-flume");
+    Properties properties = getProperties(System.getProperty(CONFIG_NAME));
+    properties.setProperty("sink1.port", "11111");
```
anew: is this why the ingestion would fail?

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+      propertyMap.put(name, properties.getProperty(name));
+    }
+    long startTime = System.currentTimeMillis();
+    agent.configure(propertyMap);
+    agent.start();
+    for (int i = 0; i < eventNumber; i++) {
+      Event event = new SimpleEvent();
+      agent.put(event);
+    }
+    agent.stop();
```
anew, line 118: how do you verify that the ingestion failed? If you don't then what's the point of this test? What are you trying to test?

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+        } catch (IOException e) {
+        }
+      }
+    }
+
+    CloseableHttpClient httpClient = HttpClients.custom().setConnectionManager(
+      new BasicHttpClientConnectionManager()).build();
+    HttpResponse response = httpClient.execute(getRequest);
+    return response;
+  }
```
anew, line 219: if this http client is closeable, then shuoldn't you close it?

anew:
@ytolochkevych please see my comments. Thanks -Andreas.

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+    String authPath = properties.getProperty(AUTH_PROPERTIES);
+    if (authPath != null) {
+      URL url = Thread.currentThread().getContextClassLoader().getResource(authPath);
+      properties.setProperty(AUTH_PROPERTIES, url.getPath());
+    }
+    Map propertyMap = new HashMap<String, String>();
+    for (final String name : properties.stringPropertyNames()) {
+      propertyMap.put(name, properties.getProperty(name));
+    }
+    long startTime = System.currentTimeMillis();
```
anwar6953: Remove if never used (currently unused).

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+    URI baseUrl = URI.create(String.format("%s://%s:%s", ssl ? "https" : "http",
+                                           host, port));
+    HttpGet getRequest = new HttpGet(baseUrl.resolve(String.format("/v2/streams/%s/events?start=%s&end=%s",
+                                                                   streamId, startTime, endTime)));
+    if (authClientPropertiesPath != null) {
+      String authClientClassName = properties.getProperty(
+        "sink1.authClientClass", "co.cask.cdap.security.authentication.client.basic.BasicAuthenticationClient");
+      AuthenticationClient authClient = (AuthenticationClient) Class.forName(authClientClassName).newInstance();
+
+      Properties authProperties = new Properties();
```
anwar6953: Remove if never used (currently unused).

cdap-integration-tests/cdap-flume-java-integration-tests/pom.xml
anwar6953: Also, the two json versions are different. What is the reason for that?

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+    agent.configure(propertyMap);
+    agent.start();
+    long startTime = System.currentTimeMillis();
+    for (int i = 0; i < eventNumber; i++) {
+      Event event = new SimpleEvent();
+      event.setBody((EVENT_STR + i).getBytes());
+      agent.put(event);
+    }
+    Thread.currentThread().sleep(1000);
```
anwar6953: Can't this simply be `Thread.sleep(1000);`
`sleep` is a static method anyways.

cdap-integration-tests/cdap-flume-java-integration-tests/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+
+  <artifactId>cdap-flume-java-integration-tests</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0.0-SNAPSHOT</version>
+  <name>CDAP Flume Integration Tests</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>stream-client-java</artifactId>
```
anwar6953: artifactId of that has been renamed to 'cdap-stream-client'.
This probably works for you, because you have the old artifact name still cached in your repo.

My assumption is that the current pom will fail to build if you  do `mvn -U clean package`, since the pom is currently specifying the old name.

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
ytolochkevych: ok

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
ytolochkevych: apache.http is already used in other  cdap-ingest modules

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
ytolochkevych: yes, wrong port is set

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
ytolochkevych, line 118: According to the logic logic of the CdapSink, it throw EventDeliveryException in a case if the message was not delivered to Cdap and try resend event. This test tests  such case. If it is redundant, I can remove it.

cdap-integration-tests/pom.xml
ytolochkevych, line 38: I can create new pull request when chis PR will be merged.

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
ytolochkevych: We assume that stream already exist ( we had similar discussion for CdapSink )

anew: That is different. The sink cannot create the stream every time it starts - it assumes that the stream has been created (the sink is a library and . The test must setup the environment for the sink to work properly, that includes creating the stream.

cdap-integration-tests/cdap-flume-java-integration-tests/src/test/java/co/cask/cdap/flume/CdapFlumeIT.java
anew, line 118: I thought you would want to test that after a failure, the sink is still functional. So you would have to  simulate a transient failure and make sure the sink recovers from that. For example, CDAP is down. Sink fails to send events. After a few seconds, CDAP is back and the sink can send events again.

ytolochkevych:
@anew I updated test logic. Now the logic is next: we start send events, then we simulate cdap fail by setting incorrect port and  send some number of events. Then port is Cdap port is returned to  correct value and agent continue to send events

anew:
Thanks, this looks much better. LGTM

PR #57, opened 2014-10-03
Feature/misc fixes

anwar6953:
Make tests be able to run from any directory (by specifying full paths for files).
 Don't have a default authentication client. Rationale: If user wants an authentication client, they should configure an authentication client (set user/password on it) and then set it onto the stream config. (Also to be more consistent with java stream client)

### DIFF
--- a/cdap-stream-clients/python/test/runtests.py
+++ b/cdap-stream-clients/python/test/runtests.py
@@ -32,13 +32,12 @@ import inspect
 current_dir = os.path.dirname(
     os.path.abspath(inspect.getfile(inspect.currentframe())))
 parent_dir = os.path.dirname(current_dir)
-src_dir = parent_dir + '/cdap_stream_client'
-sys.path.insert(0, src_dir)
+sys.path.insert(0, parent_dir)
 
-from config import Config
-from serviceconnector import NotFoundError
-from streamwriter import StreamWriter
-from streamclient import StreamClient
+from cdap_stream_client import Config
+from cdap_stream_client.serviceconnector import NotFoundError
+from cdap_stream_client import StreamWriter
+from cdap_stream_client import StreamClient
 
 with mock.patch('__main__.Config.is_auth_enabled',
                         new_callable=mock.PropertyMock) \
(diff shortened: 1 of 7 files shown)


anwar6953:
This depends on:
https://github.com/caskdata/cdap-clients/pull/36

cdap-stream-clients/python/test/runtests.py
```diff
@@ -32,13 +32,12 @@
...
+sys.path.insert(0, parent_dir)
 
-from config import Config
-from serviceconnector import NotFoundError
-from streamwriter import StreamWriter
-from streamclient import StreamClient
+from cdap_stream_client import Config
+from cdap_stream_client.serviceconnector import NotFoundError
+from cdap_stream_client import StreamWriter
+from cdap_stream_client import StreamClient
```
ParthGandhi: Condense these imports.

cdap-stream-clients/python/test/stream_integration_test.py
```diff
@@ -21,15 +21,14 @@
     import unittest2 as unittest
 except ImportError:
     import unittest as unittest
-import requests
 
 from stream_test_base import StreamTestBase
-
+import os
```
ParthGandhi: Move this import to above `from stream_test_base import StreamTestBase` since its a std library.

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -24,12 +24,12 @@
...
-from streamwriter import StreamWriter
-from streamclient import StreamClient
-from cdap_auth_client.BasicAuthenticationClient import BasicAuthenticationClient
-from cdap_auth_client.Config import Config as AuthConfig
+from cdap_stream_client import Config
+from cdap_stream_client.serviceconnector import NotFoundError
+from cdap_stream_client import StreamWriter
+from cdap_stream_client import StreamClient
+from cdap_auth_client import BasicAuthenticationClient
+from cdap_auth_client import Config as AuthConfig
```
ParthGandhi: Condense here too.

ParthGandhi:
Few comments.

ParthGandhi:
:+1:

PR #63, opened 2014-10-07
remove ruby stream client from release 1.0.0. It only recently was integ...

anwar6953:
Remove ruby stream client from release 1.0.0. It only recently was integrated with ruby auth client, and has not been rigorously tested.

Will remain on develop, for next release.


poornachandra:
LGTM :+1: 

Please merge this into develop also, and make sure that ruby client does not get removed from develop.

PR #64, opened 2014-10-07
Feature/develop keep ruby

anwar6953:
feature/develop-keep-ruby is a branch that has release merged into it. (The difference between feature/develop-keep-ruby and release/1.0.0 is that the commit to remove ruby project has been reverted in the feature/develop-keep-ruby branch.


poornachandra:
LGTM :+1:

PR #65, opened 2014-10-07
Fix dependency by using the right artifact id

sreevatsanraman:


poornachandra:
LGTM :+1:

anwar6953:
Lets change this too, in the same PR:
https://github.com/caskdata/cdap-ingest/blob/dce9abe43ca850f86de6f90b435efa7cb7c84c43/cdap-stream-clients/java/pom.xml#L52

anwar6953:
:+1:

PR #71, opened 2014-10-09
Feature/disablecertcheck verifysslcert

anwar6953:
Change disableCertCheck property to verify.ssl.cert, because when that property is set to true, it actually does the verification. Setting the property to false avoids the cert verification.

This is also now consistent with authentication client.

### DIFF
--- a/cdap-integration-tests/cdap-file-drop-zone-integration-tests/src/test/resources/remote_auth_properties.conf
+++ b/cdap-integration-tests/cdap-file-drop-zone-integration-tests/src/test/resources/remote_auth_properties.conf
@@ -2,5 +2,5 @@
 security.auth.client.username=admin
 # User password
 security.auth.client.password=realtime
-# Disable certificate validation
+# Perform certificate validation
 security.auth.client.verify.ssl.cert=true
--- a/cdap-integration-tests/cdap-file-drop-zone-integration-tests/src/test/resources/remote_auth_properties.conf
+++ b/cdap-integration-tests/cdap-file-drop-zone-integration-tests/src/test/resources/remote_auth_properties.conf
@@ -3,4 +3,4 @@ security.auth.client.username=admin
 # User password
 security.auth.client.password=realtime
 # Disable certificate validation
-security.auth.disable.certificate.check=true
+security.auth.client.verify.ssl.cert=true
--- a/cdap-integration-tests/cdap-file-drop-zone-integration-tests/src/test/resources/remote_auth_properties.conf
+++ b/cdap-integration-tests/cdap-file-drop-zone-integration-tests/src/test/resources/remote_auth_properties.conf
@@ -3,4 +3,4 @@ security.auth.client.username=admin
 # User password
 security.auth.client.password=realtime
 # Perform certificate validation
-security.auth.client.verify.ssl.cert=true
+security.auth.client.verify.ssl.cert=false
--- a/cdap-integration-tests/cdap-file-tailer-integration-tests/src/test/resources/remote_auth.conf
+++ b/cdap-integration-tests/cdap-file-tailer-integration-tests/src/test/resources/remote_auth.conf
@@ -21,6 +21,6 @@ pipes.pipe1.sink.port=10443
 # Host ssl mode that is used by stream client
 pipes.pipe1.sink.ssl=true
 # Host ssl cert check mode that is used by stream client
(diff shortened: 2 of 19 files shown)


cdap-integration-tests/cdap-file-drop-zone-integration-tests/src/test/resources/remote_auth_properties.conf
```diff
@@ -2,5 +2,5 @@
 security.auth.client.username=admin
 # User password
 security.auth.client.password=realtime
-# Disable certificate validation
-security.auth.disable.certificate.check=true
\ No newline at end of file
+# Perform certificate validation
+security.auth.client.verify.ssl.cert=true
```
poornachandra: This should be false, right? Otherwise tests will fail.

cdap-integration-tests/cdap-file-tailer-integration-tests/src/test/resources/remote_auth.conf
```diff
@@ -21,6 +21,6 @@ pipes.pipe1.sink.port=10443
 # Host ssl mode that is used by stream client
 pipes.pipe1.sink.ssl=true
 # Host ssl cert check mode that is used by stream client
-pipes.pipe1.sink.disableCertCheck=true
+pipes.pipe1.sink.verify.ssl.cert=true
```
poornachandra: This should be false, right? Otherwise tests will fail.

cdap-integration-tests/cdap-file-tailer-integration-tests/src/test/resources/remote_auth_properties.conf
```diff
@@ -2,5 +2,5 @@
 security.auth.client.username=admin
 # User password
 security.auth.client.password=realtime
-# Disable certificate validation
-security.auth.disable.certificate.check=true
\ No newline at end of file
+# Perform certificate validation
+security.auth.client.verify.ssl.cert=true
```
poornachandra: This should be false, right? Otherwise tests will fail.

cdap-integration-tests/cdap-stream-client-integration-tests/src/test/resources/remote_auth.conf
```diff
@@ -1,7 +1,7 @@
 host=remotehost
 port=10443
 ssl=true
-disableCertCheck=true
+verify.ssl.cert=true
```
poornachandra: This should be false, right? Otherwise tests will fail.

cdap-integration-tests/cdap-stream-client-integration-tests/src/test/resources/remote_auth_properties.conf
```diff
@@ -1,3 +1,3 @@
 security.auth.client.username=user
 security.auth.client.password=secret
-security.auth.disable.certificate.check=true
\ No newline at end of file
+security.auth.client.verify.ssl.cert=true
```
poornachandra: This should be false, right? Otherwise tests will fail.

cdap-stream-clients/python/test/cdap_config.json
```diff
@@ -4,5 +4,5 @@
     "SSL": false,
     "security_auth_client_username": "username",
     "security_auth_client_password": "password",
-    "security_ssl_cert_check": false
+    "security_verify_ssl_cert": true
 }
```
poornachandra, line 8: This should be false, right? Otherwise tests will fail.

cdap-stream-clients/python/test/cdap_ssl_config.json
```diff
@@ -4,5 +4,5 @@
     "SSL": true,
     "security_auth_client_username": "username",
     "security_auth_client_password": "password",
-    "security_ssl_cert_check": false
+    "security_verify_ssl_cert": true
```
poornachandra: This should be false, right? Otherwise tests will fail.

cdap-stream-clients/python/test/default-config.json
```diff
@@ -4,5 +4,5 @@
     "SSL": false,
     "security_auth_client_username": "username",
     "security_auth_client_password": "password",
-    "security_ssl_cert_check": false
+    "security_verify_ssl_cert": true
```
poornachandra: This should be false, right? Otherwise tests will fail.

poornachandra:
LGTM :+1:

PR #72, opened 2014-10-09
checking if ssl is enabled before trying to load auth client properties

gsps1:
just added a line to check if SSL is enabled before loading auth client properties.

### DIFF
--- a/cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
+++ b/cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
@@ -170,6 +170,7 @@ public class StreamSink implements Sink, LifecycleAware, Configurable {
       InputStream inStream = null;
       try {
         authClient = (AuthenticationClient) Class.forName(authClientClassName).newInstance();
+        authClient.setConnectionInfo(host, port, sslEnabled);
         if (authClient.isAuthEnabled()) {
           Properties properties = new Properties();
           properties.setProperty(BasicAuthenticationClient.VERIFY_SSL_CERT_PROP_NAME, String.valueOf(verifySSLCert));
@@ -177,7 +178,6 @@ public class StreamSink implements Sink, LifecycleAware, Configurable {
           properties.load(inStream);
           authClient.configure(properties);
         }
-        authClient.setConnectionInfo(host, port, sslEnabled);
         builder.authClient(authClient);
       } catch (IOException e) {
         LOG.error("Cannot load properties", e);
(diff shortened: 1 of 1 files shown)


anwar6953:
I don't think this is a good idea, because I think we support authentication even without ssl.

anwar6953:
You can do:

```
authClient = (AuthenticationClient) Class.forName(authClientClassName).newInstance();
authClient.setConnectionInfo(host, port, sslEnabled);
boolean authEnabled = authClient.isAuthEnabled();

if (authEnabled) {
  // load properties
}
```

Source: README: https://github.com/caskdata/cdap-clients/tree/release/1.0.0/cdap-authentication-clients/java

anwar6953:
Does it make sense to apply the same fix to the following file?
It's the same issue and same fix there.
cdap-ingest / cdap-file-tailer / src / main / java / co / cask / cdap / filetailer / config / PipeConfigurationImpl.java

Edit: Its very simple, so I went ahead and did it in:
https://github.com/caskdata/cdap-ingest/pull/74
Also tested, to make sure this problem no longer exists (in file-tailer or file-drop-zone).

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -170,13 +170,14 @@ private void createStreamClient() throws IOException {
...
-        Properties properties = new Properties();
-        properties.setProperty(BasicAuthenticationClient.VERIFY_SSL_CERT_PROP_NAME, String.valueOf(verifySSLCert));
-        inStream = new FileInputStream(authClientPropertiesPath);
-        properties.load(inStream);
-        authClient.configure(properties);
         authClient.setConnectionInfo(host, port, sslEnabled);
+        if (authClient.isAuthEnabled()) {
+          Properties properties = new Properties();
+          properties.setProperty(BasicAuthenticationClient.VERIFY_SSL_CERT_PROP_NAME, String.valueOf(verifySSLCert));
+          inStream = new FileInputStream(authClientPropertiesPath);
```
poornachandra, line 177: It would be a good idea to add a check - authClientPropertiesPath is not null or empty.

poornachandra:
One minor comment, please fix and merge :+1:

poornachandra:
This PR is against develop instead of release/1.0.0. I'm closing this PR, please open a new PR against release/1.0.0.

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
anwar6953, line 177: For file-tailer, when you install, it places an authClientProperties file in the same place as the other properties file, and sets that as the default authClientPropertiesPath. That would be a good idea here too, right?
https://github.com/caskdata/cdap-ingest/blob/31f19a16287d6e427ab30f742b9cd08b5d24b890/cdap-file-tailer/src/main/java/co/cask/cdap/filetailer/config/PipeConfigurationImpl.java#L216

poornachandra, line 177: I don't think Flume Client has any rpms or debs for installation.

PR #73, opened 2014-10-09
Remove default auth client, make config

anwar6953:
Don't have a default authentication client. Rationale: If user wants an authentication client, they would have to configure an authentication client (set user/password on it) and then set it onto the stream config. (Also to be more consistent with java stream client).

Also make `Config#read_from_file` allow a file that doesn't have all fields (optional fields).

This works with: https://github.com/caskdata/cdap-clients/pull/50.

### DIFF
--- a/cdap-stream-clients/python/cdap_stream_client/config.py
+++ b/cdap-stream-clients/python/cdap_stream_client/config.py
@@ -71,10 +71,7 @@ class Config(object):
     def auth_token(self):
         if self.__authClient is None:
             raise AttributeError("Authentication Client is not set.")
-        try:
-            return self.__authClient.get_access_token()
-        except IOError:
-            return u''
+        return self.__authClient.get_access_token()
 
     @property
     def is_auth_enabled(self):
--- a/cdap-stream-clients/python/cdap_stream_client/config.py
+++ b/cdap-stream-clients/python/cdap_stream_client/config.py
@@ -17,23 +17,16 @@
 from __future__ import with_statement
 import json
 from io import open
-from cdap_auth_client import BasicAuthenticationClient
-from cdap_auth_client import Config as AuthConfig
 
 
 class Config(object):
 
-    def __init__(self, host=u'localhost', port=10000, ssl=False,
-                 verify_ssl_cert=True, filename=u''):
+    def __init__(self, host=u'localhost', port=10000, ssl=False, verify_ssl_cert=True):
         self.__host = host
         self.__port = port
         self.__ssl = ssl
         self.__verify_ssl_cert = verify_ssl_cert
-        self.__authClient = BasicAuthenticationClient()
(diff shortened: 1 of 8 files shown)


ParthGandhi:
Does the documentation reflect that certain properties are optional (and what their default values are) when reading the file? 

Barring that, code LGTM.

ParthGandhi:
Ok, discussed more changes offline to make it cleaner.

cdap-stream-clients/python/cdap_stream_client/config.py
```diff
@@ -72,6 +65,8 @@ def ssl_cert_check(self, state):
 
     @property
     def auth_token(self):
+        if self.__authClient is None:
+            raise AttributeError("Authentication Client is not set.")
         try:
             return self.__authClient.get_access_token()
         except IOError:
```
gsps1: not part of this PR, but should we print the stack trace instead of returning empty string ?

cdap-stream-clients/python/cdap_stream_client/config.py
```diff
@@ -17,23 +17,16 @@
...
 from io import open
-from cdap_auth_client import BasicAuthenticationClient
-from cdap_auth_client import Config as AuthConfig
 
 
 class Config(object):
 
-    def __init__(self, host=u'localhost', port=10000, ssl=False,
-                 verify_ssl_cert=True, filename=u''):
+    def __init__(self, host=u'localhost', port=10000, ssl=False, verify_ssl_cert=True):
```
gsps1: may be better to have these defaults defined separately ?

cdap-stream-clients/python/cdap_stream_client/config.py
anwar6953: Changed to remove the `except`. This type of exception should not be caught. IF the get_access_token fails, we should not continue to use an empty string in further connections as the access token...

poornachandra:
LGTM once @ParthGandhi has verified the python part.

cdap-stream-clients/python/cdap_stream_client/config.py
```diff
@@ -15,25 +15,22 @@
...
+DEFAULT_PORT = 10000
+DEFAULT_SSL = False
+DEFAULT_VERIFY_SSL_CERT = True
 
 
 class Config(object):
 
-    def __init__(self, host=u'localhost', port=10000, ssl=False,
-                 verify_ssl_cert=True, filename=u''):
+    def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, ssl=DEFAULT_SSL, verify_ssl_cert=DEFAULT_VERIFY_SSL_CERT):
```
ParthGandhi: Too long.

ParthGandhi:
One comment. 

LGTM.

PR #74, opened 2014-10-11
Check if auth is enabled before trying to load auth client properties.

anwar6953:
Simply makes the following actions conditional upon `isAuthEnabled`:
-reading of auth client properties file
-configuring auth client

Solves the same problem as: https://github.com/caskdata/cdap-ingest/pull/72, but in file-tailer. (https://github.com/caskdata/cdap-ingest/pull/72 resolves the issue in flume).

Just as reference, [here](https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/PVXeULnifjRyKby/Screen%20Shot%202014-10-09%20at%2011.17.07%20PM.png) is a screenshot of logs, when omitting authclient.properties file, even if authentication is disabled in cdap, which this fix resolves.

### DIFF
--- a/cdap-file-tailer/src/main/java/co/cask/cdap/filetailer/config/PipeConfigurationImpl.java
+++ b/cdap-file-tailer/src/main/java/co/cask/cdap/filetailer/config/PipeConfigurationImpl.java
@@ -26,6 +26,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


cdap-file-tailer/src/main/java/co/cask/cdap/filetailer/config/PipeConfigurationImpl.java
```diff
@@ -272,6 +275,8 @@ public StreamClient getStreamClient() {
       } catch (ConfigurationLoadingException e) {
         LOG.warn("Can not load Authentication Client properties file {}: {}",
                   authClientPropertiesPath, e.getMessage(), e);
+      } catch (IOException e) {
+        LOG.warn("Failed to check if authorization is enabled.");
```
gsps1: can you add the exception e to the log message.

poornachandra:
LGTM :+1:

PR #75, opened 2014-10-12
In StreamWriter#write, allow empty strings, disallow null strings

anwar6953:
[CDAP-396](https://issues.cask.co/browse/CDAP-396)
In StreamWriter#write, allow empty strings and disallow null strings.
This allows for file-tailer and file-drop-zone to work with files that have empty lines.

I have verified that empty strings can be sent to cdap streams (it increases the stream's event count by 1, and data storage of the stream remains unchanged).


poornachandra:
Thanks for fixing this @anwar6953. It would be a good idea to add a test case to Stream Client unit tests to test this. LGTM :+1:

poornachandra:
LGTM :+1:

PR #76, opened 2014-10-13
Js ingest

kpovetkin:
Send, request logic is implemented for browser and nodejs.


anwar6953:
The main comment in the previous PR is that it lacked an api exposed such as this: [StreamClient](https://raw.githubusercontent.com/caskdata/cdap-ingest/develop/cdap-stream-clients/java/src/main/java/co/cask/cdap/client/StreamClient.java) and [StreamWriter](https://raw.githubusercontent.com/caskdata/cdap-ingest/develop/cdap-stream-clients/java/src/main/java/co/cask/cdap/client/StreamWriter.java). This seems to be the case here, as well.

Additionally, please address a couple of comments from the old PR https://github.com/caskdata/cdap-ingest/pull/53 (they do apply here):
Integrate JS auth client: https://github.com/caskdata/cdap-ingest/pull/53#issuecomment-57703590
Change directory structure: https://github.com/caskdata/cdap-ingest/pull/53#issuecomment-57898624
Annotation comment/question: https://github.com/caskdata/cdap-ingest/pull/53#discussion_r18428265
Error Throwing: https://github.com/caskdata/cdap-ingest/pull/53#discussion_r18428323

Thanks :)

kpovetkin:
1. Integration is in progres with other features.
2. Will change.
3. The reason why 'any' is in brackets: http://usejsdoc.org/tags-param.html .
4. Truth to be told, I've not found any 'throw' statement without 'new'. Will take closer look on this case.

anwar6953:
1. Thanks.
2. Thanks.
3. Right, but the proper annotation (even according to that page) is likely:`@param {*} somebody - Whatever you want.`
4. Search 'throw' on this page: https://github.com/caskdata/cdap-ingest/pull/76/files. There are 5 such occurrences.

Thanks!

stream-client/stream-client-javascript/src/utils.js
```diff
@@ -0,0 +1,72 @@
...
+                }
+            };
+
+            addToResult(src1);
+            addToResult(src2);
+
+            return result;
+        },
+        baseUrl: function baseUrl(hostname, port, ssl) {
+            return ['', (ssl ? 'https' : 'http'), '://',
```
anwar6953: Is the first element in this array necessary?
If not, let's remove it for simplicity.

anwar6953:
I was unable to try out the code. Trying to include the files simply returns an empty dictionary.
What am I doing wrong?

```
$ node
> require("./serviceconnector.js")
{}
```

kpovetkin:
https://github.com/caskdata/cdap-ingest/pull/76#commitcomment-8149863

This is module definition block, so it should not be extracted to any function. I'll play with build machine to remove this code to other place.

kpovetkin:
Library is completely implemented( with Authentication Manager integration ).

Docs, test and built system will be added later.

anwar6953:
Hi.

The following comment is not a big task, but you never got around to:
https://github.com/caskdata/cdap-ingest/pull/53#issuecomment-57898624

I will review the remainder of the code now.

stream-client/stream-client-javascript/src/streamclient.js
```diff
@@ -0,0 +1,223 @@
...
+    var prepareUri = function prepareUri(params) {
+            params.placeholder = params.placeholder || 'StreamID';
+            params.data = params.data || '';
+
+            return REQUESTS[params.request].replace(REQUEST_PLACEHOLDERS[params.placeholder], params.data);
+        },
+        checkErrorResponse = function checkErrorResponse(response) {
+            if (200 !== response.status) {
+                throw {
+                    status: response.status,
```
anwar6953: an object being thrown should have: `name` and `message` fields.
Otherwise, the log in the browser or node is not very useful.

anwar6953:
Edit: Looks resolved.
getTTL is ok, but setTTL doesn't seem to be working:
image

This can be fixed by changing streamclient.js (line) 139 from:
`body: objectToSend,`
to
`data: objectToSend,`
and fixing this: https://github.com/caskdata/cdap-ingest/pull/76#issuecomment-59603622

For the same reason as described in:
https://github.com/caskdata/cdap-ingest/pull/76#issuecomment-59602443

anwar6953:
Edit: Looks resolved.
In StreamWriter.js, line 69, it is sending the message as the object's `body` field.
However, in request-node.js, line 30, it is expecting it to be on the object's `data` field.

This inconsistency results in a bug that causes streamWriter.write("message") to always fail to send the message.

anwar6953:
Edit: Looks resolved.
Additionally, the `config` parameter's data/body parameter is not being copied in `requestImpl` in serviceconnector.js (only `method`, `path`, and `headers` are being copied).
Without passing the data/body parameter, a lot of important requests can't happen (any request with a body/data).

anwar6953:
Hey, I left a few comments for improvement.
If there's any questions (or if something's not clear), let me know.

Thanks.

kpovetkin:
Everythink works how it expecting to do. Unit tests added to verify. 
Documentation added.

kpovetkin:
Integration tests added.

kpovetkin:
Directory renamed

stream-client/javascript/src/streamclient.js
```diff
@@ -0,0 +1,230 @@
...
+
+            return REQUESTS[params.request].replace(REQUEST_PLACEHOLDERS[params.placeholder], params.data);
+        },
+        checkErrorResponse = function checkErrorResponse(response) {
+            var status = ('status' in response) ? response.status : response.statusCode,
+                responseText = ('responseText' in response) ? response.responseText : response.body.toString();
+
+            if (200 !== status) {
+                throw {
+                    status: status,
```
anwar6953: What I had meant in a previous comment, is that one of the object's fields should be `name`. Otherwise, when the exception is thrown, it merely shows up as [Object object], without the `message` field even printed. If you give this object a `name` field, the proper message will also be printed when the exception is thrown.

So:
`throw {
    name: ... ,
    message: ...
}`

anwar6953:
Please also rename the outer directory to: 'cdap-stream-clients', as per: https://github.com/caskdata/cdap-ingest/pull/53#issuecomment-57898624

stream-client/javascript/src/request-node.js
```diff
@@ -0,0 +1,161 @@
...
+var requestAsync = function requestAsync(params) {
+        var connection = params.ssl ? require('https') : require('http');
+
+        if (!params.method || !params.path || !params.host || !params.port) {
+            throw new Error('"host", "port", "method", "path" properties are required');
+        }
+
+        params.ssl = (null != params.ssl) ? params.ssl : false;
+        params.headers = params.headers || {};
+        params.data = JSON.stringify(params.data || {});
```
anwar6953: In all of the other clients, when a user does StreamWriter.write("11byteslong"), simply 11 bytes get sent in the data field. However, here, 13 bytes are being sent because a string is being preceded and followed by escaped quotes, when it is `stringified`.

To avoid that, don't stringify the data parameter, and expect it as a string, rather than an object. This change also makes it consistent with the StreamWriters in the other languages.

You can use this endpoint to see which events are in the stream:
http://localhost:10000/v2/streams/{streamName}/events

kpovetkin: Changed 

----- Original Message -----

stream-client/javascript/Build.md
```diff
@@ -0,0 +1,37 @@
+# Build procedure.
+
+Build process is implemented with [Grunt](http://gruntjs.com/).
+
+to set up all dependencies use next shell commands:
+
+```
+# npm install -g grunt-cli
+# npm unstall -g bower
```
anwar6953: unstall -> install

anwar6953:
When I run `grunt`, some tests fail.
[Here's a screenshot.](https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/gW7IAWmjuWkIDhd/Screen%20Shot%202014-10-27%20at%208.40.05%20AM%201.png)

This is after running the following steps
$ npm install -g grunt-cli
$ npm install -g bower
$ npm install
$ bower install

My node version is v0.10.29.

anwar6953:
It looks like the integration tests are running off a separate copy of the code (instead of the code in /src directory).
So, whenever changes are made to /src files, the integration tests are running against a stale copy of the code (unless we manually copy the files over).

Can we either:
1) run the tests against the source code in /src directory
            -or-
2) if that's not possible, then update the copy of the code that the integration tests run against, as a part of the test.

anwar6953:
Some other places were affected by the change in this commit, so you have to change them too.
https://github.com/cybervisiontech/cdap-ingest/commit/0ed09d67079fd13a0e966e078831f08b943937d4

For instance, stream client's [setTTL method's implementation](https://github.com/caskdata/cdap-ingest/pull/76/files#diff-9e02088df7f5c168f408090a011b99e7R133) relied on the data being JSON.stringified, which isn't being done now, so the JSON.stringify call probably has to be done in the setTTL method.

stream-client/javascript/src/streamclient.js
kpovetkin: No, it would not, at least not in most cases. The better way is to use toString method.

kpovetkin:
Updated according to comments.

kpovetkin:
This PR should be replaced by this one: https://github.com/caskdata/cdap-ingest/pull/93

PR #77, opened 2014-10-13
adding auth client isEnabled check before loading the properties file fo...

gsps1:
...r authentication client configuration

based of , https://github.com/caskdata/cdap-ingest/pull/72, opening against release/1.0.0 as per @poornachandra suggestion.

### DIFF
--- a/cdap-flume/README.md
+++ b/cdap-flume/README.md
@@ -51,6 +51,11 @@ Path to authentication client properties file:
 
     a1.sinks.sink1.authClientProperties = /usr/local/apache-flume/conf/auth_client.conf
 
+Optional Security configuration:
+
+    a1.sinks.sink1.sslEnabled = false (set true to use HTTPS)
+    a1.sinks.sink1.verifySSLCert = true (set false to suspend certificate checks; this allows self-signed certificates when SSL is true)
+     
 ### Authentication Client Example Configuration
 
     # User name
@@ -76,3 +81,4 @@ Configuration of a Flume agent that reads data from a log file and puts it to CD
     a1.channels.c1.type = memory
     a1.channels.c1.capacity = 1000
     a1.channels.c1.transactionCapacity = 100
+ 
--- a/cdap-flume/README.md
+++ b/cdap-flume/README.md
@@ -28,7 +28,11 @@ Enter the host port that is used by the Stream Client:
 
 Secure Socket Layer mode [true | false]
 
-    a1.sinks.sink1.sslEnabled = false 
+    a1.sinks.sink1.sslEnabled = false
+
+Verify SSL Certificate [true | false]
+    
+    a1.sinks.sink1.verifySSLCert = true (set false to suspend certificate checks; this allows self-signed certificates when SSL is true) 
 
 Number of threads to which Stream Client can send events:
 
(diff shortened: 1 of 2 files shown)


cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -170,13 +170,17 @@ private void createStreamClient() throws IOException {
...
-        inStream = new FileInputStream(authClientPropertiesPath);
-        properties.load(inStream);
-        authClient.configure(properties);
         authClient.setConnectionInfo(host, port, sslEnabled);
+        if (authClient.isAuthEnabled()) {
+          Properties properties = new Properties();
+          properties.setProperty(BasicAuthenticationClient.VERIFY_SSL_CERT_PROP_NAME, String.valueOf(verifySSLCert));
+          if ((authClientPropertiesPath == null) || (authClientPropertiesPath.isEmpty())) {
+            LOG.error("Authentication client is enabled, but the path for properties file is either empty or null");
+          }
```
anwar6953, line 187: Should we wrap the next three lines in an `else`, so they are not executed in the case that authClientPropertiesPath is not valid?

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -170,13 +170,17 @@ private void createStreamClient() throws IOException {
...
+          inStream = new FileInputStream(authClientPropertiesPath);
```
chtyim: stream not closed after use.

gsps1: its closed in finally , didn't come up in diff

cdap-flume/src/main/java/co/cask/cdap/flume/StreamSink.java
```diff
@@ -170,18 +171,28 @@ private void createStreamClient() throws IOException {
...
-        authClient.configure(properties);
         authClient.setConnectionInfo(host, port, sslEnabled);
+        if (authClient.isAuthEnabled()) {
+          Properties properties = new Properties();
+          properties.setProperty(BasicAuthenticationClient.VERIFY_SSL_CERT_PROP_NAME, String.valueOf(verifySSLCert));
+          if ((authClientPropertiesPath == null) || (authClientPropertiesPath.isEmpty())) {
+            throw new Exception("Authentication client is enabled, but the path for properties " +
+                                  "file is either empty or null");
+          }
+          inStream = new FileInputStream(authClientPropertiesPath);
```
chtyim: Should just do the close using finally in here. It makes the code much simplier.

``` java
InputStream is = new FileInputStream(...);
try {
  properties.load(is);
} finally {
  Closeables.closeQuietly(is);
}
```

cdap-flume/README.md
```diff
@@ -51,6 +51,11 @@ Path to authentication client properties file:
 
     a1.sinks.sink1.authClientProperties = /usr/local/apache-flume/conf/auth_client.conf
 
+Optional Security configuration:
+
+    a1.sinks.sink1.sslEnabled = false (set true to use HTTPS)
```
anwar6953: This one line (56) seems to already be documented at line 31 of this file.
The verifySSLCert doc addition (line 57) is good though.

poornachandra: Also, a1.sinks.sink1.verifySSLCert should go right after line 31 documentation of a1.sinks.sink1.sslEnabled, as they are related.

poornachandra:
LGTM :+1:

cdap-flume/README.md
```diff
@@ -28,7 +28,11 @@ Enter the host port that is used by the Stream Client:
 
 Secure Socket Layer mode [true | false]
 
-    a1.sinks.sink1.sslEnabled = false 
+    a1.sinks.sink1.sslEnabled = false
+
+Verify SSL Certificate [true | false]
+    
+    a1.sinks.sink1.verifySSLCert = true (set false to suspend certificate checks; this allows self-signed certificates when SSL is true)
```
awholegunch, line 35: This isn't going to look very good; it really should be:

```
a1.sinks.sink1.verifySSLCert = true
```

Set false to suspend certificate checks; this allows self-signed certificates when SSL is true.

PR #78, opened 2014-10-13
Update artifactId in README

anwar6953:
This occurrence was missed when the artifactId was changed previously.


poornachandra:
LGTM :+1:

PR #79, opened 2014-10-14
install pip in README

anwar6953:
Setup instructions instruct user to use pip, instead of downloading and setting up manually.


cdap-stream-clients/python/README.md
```diff
@@ -12,18 +12,10 @@ The Stream Client Python API is for managing Streams from Python applications.
```
ParthGandhi, line 12: Remove semicolons from all the python code in this file.

ParthGandhi:
LGTM.

PR #89, opened 2014-10-23
read readme in setup.py, for rending on pypi

anwar6953:
It would look something like this (after the contents of the README.md have been translated to rst format):
https://testpypi.python.org/pypi/cdap-stream-client/1.0.1

Note: PyPi renders rst text, not md text. So, the readme will not be properly formatted on PyPi if it is in md format. PanDoc can be used to translate the format, but maybe it would be better to simply replace the current README.md with a README.rst, so we don't have to do this translation often.

Optionally, we can simply link to our github repo for the README, instead of having it both on PyPi and GitHub.

EDIT:
Resolution is to simply link to the README on github as Package Documentation on PyPi
https://pypi.python.org/pypi/cdap-stream-client


ParthGandhi:
"So, the readme will not be properly formatted on PyPi if it is not in md format". I assume you mean "... if it is in md format" (since thats what the previous line implies). 

Formatting looks good to me right now. Whats the problem?

anwar6953:
Yes, I meant to say: "So, the readme will not be properly formatted on PyPi if it is in md format." I edited original post accordingly.

The problem is that I had to translate the README using pandoc -- which is simple, but this becomes an additional step in the release process.

ParthGandhi:
I see. 

Assuming we want to stick with .md files in our Github repos, I think we shouldn't introduce this extra step. Odds are we'll forget it during some release unless we automate it. And while it is pretty cool, I don't think it adds _that_ much more benefit to warrant an additional step (since the docs are already hosted elsewhere).

@poornachandra ?

ParthGandhi:
Ok, how about this: When you run setup.py to push to PyPi, we dynamically create a .rst file from the current .md file (using https://pypi.python.org/pypi/pypandoc/), and call `setup` with a `long_description`. 

On all other runs of setup.py, we simply don't add the `long_description` field. 

This way, we can maintain only one version of the docs. And only the person pushing to PyPi needs to have pypandoc installed (not everyone using it).

poornachandra:
I see that we are directly using README.md for long_description in this PR, does it mean that the README will not be formatted properly? If so - why are we making this change (adding md file) then?

As @ParthGandhi suggested, either dynamically convert md to rst. Or as you suggested, just have a link to the github README.

ParthGandhi:
As I understand it, he locally changed the contents of his .md file to rst (without renaming or committing it), and pushed it to testpypi to just show what it would look like.

poornachandra:
btw - there is a PR open to support md format in pypi - https://bitbucket.org/pypa/pypi/pull-request/33/support-markdown-for-readmes-fixes-148/diff

ParthGandhi:
Its 6 months old, and looking at the comments - https://bitbucket.org/pypa/pypi/issue/148/support-markdown-for-readmes - i'm not sure it'll be merged anytime soon (though it did get 1 approval last month, so things might have changed).

I'm ok with just linking to github for now, and merging this fix when that PR gets merged.

ParthGandhi:
LGTM.

ParthGandhi:
Just noticed that neither README talks about which versions of Python are supported. Could you add that?

cdap-stream-clients/python/README.md
```diff
@@ -23,6 +23,7 @@ To install the development version, clone the repository
     $ cd cdap-ingest/cdap-stream-clients/python/
     $ python setup.py install
 
+Supported Python versions: 2.6, 2.7.
```
ParthGandhi: Looks better without the period at the end.

ParthGandhi:
One comment. Fix and merge.

PR #90, opened 2014-10-29
CDAP-492: Update Java Stream Client integration tests.

amakogoncv:
Updated Java Stream Client integration tests according to the issue description [CDAP-488](https://issues.cask.co/browse/CDAP-488). @anwar6953 , please take a look.


cdap-integration-tests/cdap-stream-client-integration-tests/src/test/java/co/cask/cdap/client/StreamClientIT.java
```diff
@@ -16,73 +16,164 @@
...
+      writer.close();
+    }
+    List<JsonObject> streamEvents = streamClientTestHelper.getStreamEvents(TEST_STREAM);
+    Assert.assertEquals(expectedEventsNum, streamEvents.size());
 
     //Test that we are able to truncate stream
-    client.truncate(TEST_STREAM);
+    streamClient.truncate(TEST_STREAM);
+    streamEvents = streamClientTestHelper.getStreamEvents(TEST_STREAM);
+    Assert.assertEquals(0, streamEvents.size());
```
anwar6953, line 107: I think it will be good to check the returned contents (to ensure they match with what we sent), in addition to the checking just that the size matches.

anwar6953:
Most of the test cases are good now, but one concern:

I included a statement that should fail the integration test into one of the test cases:
`Assert.assertEquals(1, 2);`

However, when I run this command to run the test:
`mvn package -P it-local`
it still shows BUILD SUCCESS at the end.

This isn't proper, right.

amakogoncv:
I tried to do the same and got the same result, but I noticed that IT were not run. Please, try to run with such command: 
`mvm -rf crap-integration-tests clean install -P it-local`

For me this works ok and I got assertion error, as expected with : 
`Assert.assertEquals(1, 2);`

cdap-integration-tests/cdap-stream-client-integration-tests/src/test/java/co/cask/cdap/client/StreamClientIT.java
amakogoncv, line 107: will done

anwar6953:
Yea, you're right regarding the integration test not running.
Turns out the test will run for 'clean install', but not for 'clean package':
`mvn -rf cdap-integration-tests clean package -P it-local`

anwar6953:
Mostly looks good to me.

Though I have one question:
In this PR: https://github.com/caskdata/cdap-ingest/pull/94
there is a utility StreamReader. Would it be better to also use that here? Or to use the streamReader in here over there? (to avoid redundancy)

amakogoncv:
Yes, `StreamReader` could be reuse here.

amakogoncv:
@anwar6953 , I have updated `StreamReader` for adapt it for all integration tests. Then I have updated Stream Client integration tests. So, for submit this PR, https://github.com/caskdata/cdap-ingest/pull/94 PR should be reviewed and submitted at first.

sreevatsanraman:
@amakogoncv PR 94 is merged. Please pull in the changes and do the needful here.

sreevatsanraman:
lgtm

PR #91, opened 2014-10-30
CDAP-488 Improve integration tests for python

ytolochkevych:


anwar6953:
Please use the current caskdata/cdap-ingest develop branch as the base for these changes, so that there are no conflicts.

(Pulling the develop branch from caskdata/cdap-ingest into cybervisiontech/cdap-ingest would be the appropriate action going forward). Let me know if I'm mistaken and this is already done, but I think this is the reason for the merge conflict.

**EDIT:** cybervisiontech/cdap-ingest is already up to date with caskdata/cdap-ingest, but this branch (CDAP-488-python-it) is currently 125 commits behind cybervisiontech/cdap-ingest.

ytolochkevych:
@anwar6953 Fixed

ytolochkevych:
@kpovetkin fixed http -> https

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -13,20 +13,29 @@
 #  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #  License for the specific language governing permissions and limitations under
 #  the License.
+import json
```
ParthGandhi: You've already imported this below.

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -61,33 +77,40 @@ def read_from_file(filename):
...
+        if self.auth_client.is_auth_enabled():
+            self.auth_client.configure(auth_config)
+            self.config.set_auth_client(self.auth_client)
         self.sc = StreamClient(self.config)
 
     def test_reactor_successful_connection(self):
+
         # Create stream
         self.sc.create(self.validStream)
+        get_stream_url = '/v2/streams/%s' % self.validStream
```
ParthGandhi: `validStream` should be "_" separated.

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -35,14 +44,14 @@
...
-
     validStream = u'validStream'
     invalidStream = u'invalidStream'
 
     validFile = u'some.log'
     invalidFile = u'invalid.file'
 
-    messageToWrite = u'some message'
+    messageToWrite = u'some_message'
+    eventsNumber = 50
```
anwar6953: what is this for? (its unused)

Edit: Use this variable in place of '50' below.

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -35,14 +44,14 @@
...
 
-
     validStream = u'validStream'
     invalidStream = u'invalidStream'
 
     validFile = u'some.log'
     invalidFile = u'invalid.file'
 
-    messageToWrite = u'some message'
+    messageToWrite = u'some_message'
```
ParthGandhi: Same here. And the line below.

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -107,9 +130,28 @@ def on_response(response):
         def check_exit_code(response):
             self.assertEqual(self.exit_code, 200)
 
-        q = sw.write(self.messageToWrite)
-        q.on_response(on_response)
-        q.on_response(check_exit_code, check_exit_code)
+        start_time = int(round(time.time() * 1000))
```
ParthGandhi, line 131: You're calling `round` to potentially get the ceiling or floor of the float (as opposed to just casting, that casts to the floor), right? Thats OK, but I'm just wondering if its necessary?

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -107,9 +130,28 @@ def on_response(response):
...
+            q.on_response(on_response)
+            q.on_response(check_exit_code, check_exit_code)
+        time.sleep(1)
+        end_time = int(round(time.time() * 1000))
+        event_request_url = \
+            u'/v2/streams/%s/events?start=%s&end=%s' % (self.validStream,
+                                                        start_time, end_time)
+        received_events = self.get_data_from_cdap(event_request_url)
+        for i in xrange(50):
+            self.assertTrue(any(event['body'] ==
```
ParthGandhi: This block is confusing to read. We can precompute the expected values, the received values, and do an assert on the sets.

``` python
self.assertEqual(50, len(received_events))
expected_events = [msg + str(i) for i in xrange(50)]
received_event_bodies = [event['body'] for event in received_events]
self.assertTrue(set(expected_events) == set(received_events))
```

This also has better running time (not that it matters for tests).

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -131,3 +173,18 @@ def test_set_ttl_invalid_stream(self):
             NotFoundError,
             self.sc.create_writer,
             self.invalidStream)
+
+    def get_data_from_cdap(self, request_url):
+        base_url = u'%s://%s:%d' % ("https" if self.config.ssl else "http",
```
ParthGandhi, line 176: Good to have a "/" at the end of `base_url` in case the caller does not send in a path beginning with "/".

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -131,3 +173,18 @@ def test_set_ttl_invalid_stream(self):
...
+                                    self.config.host, self.config.port)
+        url = base_url + request_url
+        if self.auth_client.is_auth_enabled():
+            token = self.auth_client.get_access_token()
+            headers = {'Authorization': token.token_type + " " + token.value}
+            response = requests.get(url, headers=headers)
+        else:
+            response = requests.get(url)
+        if response.status_code == httplib.NO_CONTENT:
```
ParthGandhi, line 185: This doesn't really take into account that you can get a non-JSON response for non-204 responses. 

Simpler would be to just put this in a try-catch, and say that it returns `None` if the response can't be JSON decoded (i.e. on an exception of the decode). That way we can also avoid using `httplib`.

cdap-stream-clients/python/test/stream_test_base.py
ParthGandhi: Leave this here, but replace the integers used in the loops in the test with this (since thats what it was originally written for, I guess).

ParthGandhi:
Done with a review. Please address comments.

cdap-stream-clients/python/test/stream_test_base.py
ParthGandhi: In fact, a bit cleaner would be to do 

``` python
event_bodies = [msg + str(i) for i in xrange(50)]
```

on line 135, and then iterate over `event_bodies` to send the events. Then you don't need to recreate that object here.

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -107,9 +130,28 @@ def on_response(response):
...
-        q = sw.write(self.messageToWrite)
-        q.on_response(on_response)
-        q.on_response(check_exit_code, check_exit_code)
+        start_time = int(round(time.time() * 1000))
+
+        for i in xrange(50):
+            q = sw.write(self.messageToWrite + str(i))
+            q.on_response(on_response)
+            q.on_response(check_exit_code, check_exit_code)
+        time.sleep(1)
```
anwar6953: Making 50 (eventsNumber) requests to a remote CDAP instance can potentially take over 1 second.

Would it be better to use the sw.write()'s callback to determine when all of the requests are complete (and then to do the assertions below)?

cdap-stream-clients/python/test/stream_test_base.py
ParthGandhi, line 185: Ignore this comment. Its OK as is.

cdap-stream-clients/python/test/stream_test_base.py
ytolochkevych, line 131: right, It helps to get a more accurate time value

cdap-stream-clients/python/test/stream_integration_test_ssl.py
```diff
@@ -29,7 +29,11 @@
 class TestStreamClient(unittest.TestCase, StreamTestBase):
 
     def setUp(self):
-        self.config_file = u'cdap_ssl_config.json'
+
```
ParthGandhi: Extra new line?

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -14,19 +14,27 @@
...
-
-import inspect
 import json
 import os
 import sys
+import inspect
+import threading
+import time
+import httplib
+import requests
```
ParthGandhi: Don't you need a blank line between standard libs and third party libs?

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -61,55 +76,85 @@ def read_from_file(filename):
...
+            q.on_response(check_exit_code, check_exit_code)
+        # time.sleep(1)
+        self.event_latch.wait_for_complite()
+        end_time = int(round(time.time() * 1000))
+        event_request_url = \
+            u'/v2/streams/%s/events?start=%s&end=%s' % (self.valid_stream,
+                                                        start_time, end_time)
+        received_events = self.get_data_from_cdap(event_request_url)
+        self.assertEqual(self.event_number, len(received_events))
+        expected_events = [self.message_to_write + str(i) for i in xrange(50)]
```
ParthGandhi: use `event_number` here too.

anwar6953:
wait_for_complite -> wait_for_complete
The rest looks good to me. :+1: 

@ParthGandhi ?

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -61,55 +76,84 @@ def read_from_file(filename):
...
-        q.on_response(check_exit_code, check_exit_code)
+        start_time = int(round(time.time() * 1000))
+        event_bodies = [self.message_to_write + str(i)
+                        for i in xrange(self.event_number)]
+
+        for event in event_bodies:
+            q = sw.write(event)
+            q.on_response(on_response)
+            q.on_response(check_exit_code, check_exit_code)
+        # time.sleep(1)
```
ParthGandhi: Remove commented line.

ParthGandhi:
One minor comment. Otherwise LGTM. 

Thanks for the quick fixes @ytolochkevych !

ytolochkevych:
@ParthGandhi deleted. Thanks

cdap-stream-clients/python/test/stream_test_base.py
```diff
@@ -14,19 +14,27 @@
...
 parentdir = os.path.dirname(currentdir)
 sys.path.insert(0, parentdir)
 
-from cdap_stream_client import Config, StreamWriter, StreamClient
+from cdap_stream_client.config import Config
 from cdap_stream_client.serviceconnector import NotFoundError
-from cdap_auth_client import BasicAuthenticationClient
+from cdap_stream_client.streamwriter import StreamWriter
+from cdap_stream_client.streamclient import StreamClient
+from cdap_auth_client.BasicAuthenticationClient \
```
anwar6953: Any reason why you changed this import?
This new import does not work for me, but old statement worked for me:
`from cdap_auth_client import BasicAuthenticationClient`

This also works:
`from cdap_auth_client.basic_authentication_client import BasicAuthenticationClient`

but the current import in the PR does not work for me.

anwar6953: Does this work on your machine -- am I missing something?

ytolochkevych: @anwar6953 hm, on my machine vise versa, old statement is not working

anwar6953: Resolved offline. Turned out to be due to naming differences in old version of auth_client and the current version on develop.

Therefore, reverted the import changes for cdap_auth_client.

ParthGandhi:
LGTM after the import change.

PR #93, opened 2014-10-31
[CYB-3, CDAP-493] js ingest library with complete integration tests

kpovetkin:
This branch was created to resolve in more safer way merge conflicts.


cdap-stream-clients/javascript/test/integration/integration-browser.js
```diff
@@ -0,0 +1,311 @@
+var StreamClient = CDAPStreamClient.StreamClient,
+    AuthManager = CDAPAuthManager,
+
+    user = 'kpovetkin',
```
anwar6953: can we extract this information (host,port,ssl,user,pass, etc) to a config file. and the test just reads from that file.
(This way, we can easily just keep four config files (for ssl, non-ssl, auth-enabled, etc), and just run the test with a different config file, instead of modifying the source for different test types).

cdap-stream-clients/javascript/test/integration/Readme.md
```diff
@@ -0,0 +1,13 @@
...
+You should to install dependencies prior to tests running.
+See *Build.md* in the project root.
+
+## Browser
+Open file *integration-browser.html* in your browser.
+
+## NodeJS
+Run next command from the *integration* directory:
+```
+$ mocha mocha integration-node.js
```
anwar6953: Is the duplicate 'mocha' a mistake?

Otherwise:

```
Error: cannot resolve path (or pattern) 'mocha'
```

anwar6953:
I don't think a copy of authentication client sources should be in here (even if its for testing).  Otherwise, we'd be testing with a stale version of auth client.

It should simply have a dependency on auth client, which uses the locally installed version, right?

cdap-stream-clients/javascript/test/integration/Readme.md
kpovetkin: It's a mistake. Corrected.

cdap-stream-clients/javascript/test/integration/integration-browser.js
kpovetkin: Implemeted.

kpovetkin:
https://github.com/caskdata/cdap-ingest/pull/93#issuecomment-61561024
Implemented.

cdap-stream-clients/javascript/test/integration/integration-node.js
```diff
@@ -0,0 +1,318 @@
...
+            var streamName = 'newStream',
+                streamClient = new StreamClient({
+                    host: config.host,
+                    port: config.port,
+                    ssl: config.ssl,
+                    authManager: authManager
+                });
+
+            streamClient.create(streamName);
+            streamClient.truncate(streamName);
```
anwar6953, line 155: send some events to the stream, call truncate, and then verify that 0 events are returned by cdap - in order to validate that this truncate call is actually truncating the stream.

cdap-stream-clients/javascript/test/integration/integration-node.js
```diff
@@ -0,0 +1,318 @@
...
+                            method: 'GET',
+                            host: config.host,
+                            port: config.port,
+                            protocol: config.ssl ? 'https' : 'http',
+                            headers: {
+                                Authorization: [authToken.type, ' ', authToken.token].join('')
+                            },
+                            body: '',
+
+                            path: '/v2/streams/' + streamName + '/events'
```
anwar6953: is it possible to specify start/end times in this request (similar to java integration tests), so that previous runs of the integration test do not interfere with the assertions in this run?

(otherwise if the streamWriter.write() fails for one test execution, this might still pass because the event was in the stream from before the test ran).

anwar6953:
Two comments related to new integration tests added to this pr. other than that, :+1:

anwar6953:
Note: the sync vs async discussion also applies to this (also because this utilizes the auth client).
https://github.com/caskdata/cdap-clients/pull/48#issuecomment-61693253

For now, we can merge, and then work on it in a separate PR if necessary. :+1:

PR #95, opened 2014-10-31
CDAP-488: added integration tests for ruby

yg2017:
Branch is ready for review.

### DIFF
--- a/cdap-stream-clients/ruby/spec/lib/stream-client-ruby/integration_tests_spec.rb
+++ b/cdap-stream-clients/ruby/spec/lib/stream-client-ruby/integration_tests_spec.rb
@@ -14,7 +14,6 @@ describe CDAPIngest::StreamClient do
     CDAPIngest::Rest.ssl         = config['ssl']
   end
 
-  let(:rest) {CDAPIngest::Rest.new}
   let(:stream_client){ CDAPIngest::StreamClient.new }
   let(:stream) {"LogAnalyticsFlow"}
 
@@ -24,12 +23,12 @@ describe CDAPIngest::StreamClient do
     else
       url = "#{stream_id}/events?start=#{start_time}&end=#{end_time}"
     end
-    rest.request('get', url).parsed_response
+    stream_client.rest.request('get', url).parsed_response
   end
 
   def test_create_stream stream_id
     stream_client.create stream_id
-    data = rest.request('get', '').parsed_response
+    data = stream_client.rest.request('get', '').parsed_response
     found = data.find {|e| e['name'] == stream_id}
     expect(found).not_to be_nil
   end
@@ -72,7 +71,7 @@ describe CDAPIngest::StreamClient do
     expect(found).not_to be_nil
   end
 
-  it 'test client without ssl and authentication' do
+  it 'test client without ssl and authentication', :type => 'it-local', :it => true do
 
     load_config 'spec/reactor.yml'
     # test stream creating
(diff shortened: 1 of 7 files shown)


cdap-stream-clients/ruby/spec/lib/stream-client-ruby/integration_tests_spec.rb
```diff
@@ -15,16 +14,54 @@ def load_config file
...
+    stream_client.set_ttl stream_id, ttl_test_value
+    ttl = stream_client.get_ttl stream_id
+    expect(ttl).to eq ttl_test_value
+  end
+
+  def test_writer stream_id
+    writer_test_value = 'test_body'
+    writer = stream_client.create_writer stream_id
+    writer.write writer_test_value
+    data = rest.request('get', "#{stream_id}/events").parsed_response
```
anwar6953: Similar to the integration tests in java or python, we need to specify a time frame, to avoid picking up events that were sent in previous runs of the integration test. Otherwise, even if one execution of the test fails to write to stream, it may still pass because without specifying start/end times, this gets all the events in the stream's history.

Edit: an alternative would be to truncate the stream before each test run, but specifying times for start/end should be good, and will be consistent with other tests (java, python, etc).

anwar6953:
This has a dependency on cdap-authentication-client.
I tried to install that:

```
Alis-MacBook-Pro:ruby alianwar$ pwd
/Users/alianwar/dev/cdap-clients/cdap-authentication-clients/ruby
Alis-MacBook-Pro:ruby alianwar$ sudo rake install
cdap-authentication-client 1.1.0 built to pkg/cdap-authentication-client-1.1.0.gem.
cdap-authentication-client (1.1.0) installed.
```

However, the folder it deploys is empty, and so I was unable to run the test.

```
Alis-MacBook-Pro:ruby alianwar$ cd /Library/Ruby/Gems/2.0.0/gems/
Alis-MacBook-Pro:gems alianwar$ cd cdap-authentication-client-1.1.0/
Alis-MacBook-Pro:cdap-authentication-client-1.1.0 alianwar$ ls
Alis-MacBook-Pro:cdap-authentication-client-1.1.0 alianwar$ 
```

How were you able to resolve this dependency? (locally installing the package did not work for me).

yg2017:
The problem is in the cdap-authentication-client.gemspec, I will fix it.

yg2017:
The problem with dependency on cdap-authentication-client was fixed in PR https://github.com/caskdata/cdap-clients/pull/63

yg2017:
I've changed stream writer test. Please review.

anwar6953:
https://github.com/caskdata/cdap-clients/pull/63
That PR did fix the problem.

However, I am still having some unexpected test failures problems:
https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/YM8oHOir3qVQeTJ/Screen%20Shot%202014-11-04%20at%203.34.10%20PM.png

no. 4 is a unit test right, and yet it fails?
(also, no.1 and no.2 might fail because I don't have CDAP runnning with auth/ssl), but no.3 shouldn't be failing at least, right?)

yg2017:
You have weird promise.rb version.
This http://www.rubydoc.info/gems/promise.rb/0.3.0/Promise doesn't have any parameters in initialize.
I have specified the last version 0.6.1 of promise.rb in Gemfile.
Please, check it by executing "bundle list".

yg2017:
I suppose, that problem appears on Ruby 2.0.0. 
I tested on 1.9.3.

wolf31o2:
The default Ruby with Mac OS X starting with Mavericks is 2.0, so it should definitely work, there. I would say a `minimum` of 1.9.3, but definitely needs to work in Ruby 2.0+

yg2017:
The problem was caused by missing "promise.rb" gem. Fixed by running
"bundle list" and "bundle install".

2014-11-06 0:03 GMT+02:00 Chris Gianelloni notifications@github.com:

yg2017:
I have added configuration (and updated documentation) for launch tests against different CDAP security modes. Please review.
But first this PR https://github.com/caskdata/cdap-clients/pull/64 should be merged into cdap-clients.

anwar6953:
Ok, yeah the (compatiblility) issue was that the gemfile did not previously specify a version of promise, and so my machine pulled in an different version, and so it did not work.

anwar6953:
The added test configuration looks good, along with most of the PR.

One last concern though, I received a Deprecated Warning when I ran the integration test:
https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/IuliO45FSInIiqP/Screen%20Shot%202014-11-11%20at%206.28.53%20PM.png

If that is easy to resolve (use the more recent functionality), then can we do that and then merge.
Thanks.

yg2017:
@anwar6953 I have fixed Deprecated Warning. Please take a look.

anwar6953:
:+1:

PR #96, opened 2014-11-03
CDAP-389: Integration Tests of File Drop Zone pass when they should fail.

artemmakogon:
added validations for DropZone IT


sreevatsanraman:
lgtm :+1:

PR #100, opened 2014-11-20
[cdap-597] business logic update according https://issues.cask.co/browse/CDAP-597

kpovetkin:
Changes added only for business logic. Updates for tests would be added later.

### DIFF
--- a/cdap-stream-clients/javascript/Readme.md
+++ b/cdap-stream-clients/javascript/Readme.md
@@ -81,7 +81,10 @@ Optional configurations that can be set (and their default values):
  Create a ```StreamWriter``` instance for writing events to the Stream "streamName":
 
  ```
-   var streamWriter = streamClient.createWriter("streamName");
+   var streamWriterPromise = streamClient.createWriter("streamName");
+   streamWriterPromise.then(function(writer){
+       console.log('new writer instance has been created:', writer);
+   })
  ```
 
  To write new events to the Stream, you can use either of these these methods of the ```StreamWriter``` class:
@@ -96,13 +99,21 @@ Optional configurations that can be set (and their default values):
  Example:
 
  ```
-   var streamPromise = streamWriter.write("New log event");
+   var streamWriter = streamClient.createWriter("streamName");
+   streamWriter.then(function(writer){
+       writer.write("Many messages sent to stream.");
+       writer.write("Another message sent to stream.");
+       writer.write("Multiple messages sent to stream.");
+   })
  ```
 
  To truncate the Stream *streamName*, use:
 
  ```
-   streamClient.truncate("streamName");
+   var truncatePromise = streamClient.truncate("streamName");
+   truncatePromise.then(function () {
+       console.log('truncated successfully');
+   });
  ```
 
  ### StreamPromise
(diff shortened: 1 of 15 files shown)


kpovetkin:
Unit tests updated.

kpovetkin:
There is a problem with library implementation in a way asked in the [PR](https://issues.cask.co/browse/CDAP-597). Library becomes inconsistent with Java, Python ones. 
The difference is in error handling. Promises usage makes exception usage impossible. 
The new way for error handling is to use promise`s 'catch' method.

kpovetkin:
Known issue:
 Browser tests work _only_ if html file is opened via browser. In this case tests errors could be seen in browser console. 

Requires time for the investigation.

kpovetkin:
Integration tests have been updated for nodejs _only_. Integration tests for browser will updated when issue mentioned above would be resolved.

kpovetkin:
Browser tests improved. Tests run correctly using browser or 'grunt' command.

anwar6953:
Seems like request-node.js is still using the http-sync library?

```
alis-mbp:src alianwar$ grep -r http-sync .
./nodejs/package.json:    "http-sync": "0.0.5"
./request-node.js:        var httpSync = require('http-sync');
```

anwar6953:
shouldn't the flags also be removed, which determined whether async or sync?
(since async: true is always set).

anwar6953:
One thing I ran into while using this: I was unable to send an empty string.

In Streamwriter:
`if (!message) {
                throw new Error('"message" is required');
            }`
Instead, I think it should check if `message === undefined`, then throw the Error (And allow empty string).

kpovetkin:
https://github.com/caskdata/cdap-ingest/pull/100#issuecomment-64174733

Empty message is the same as undefined message. Is there any reason to send an empty one?

anwar6953:
I don't think the two are the same.

$ emptyString = "";
""
$ emptyString === undefined
false
$ emptyString == undefined
false

And one use case where empty strings would be sent is if a file was being read line by line, and each line was sent as a message. If there's an empty line, it shouldn't throw an error, right?

kpovetkin:
I'm sorry I've muddled up undefined and boolean. Updated.

anwar6953:
Also, the README is outdated.

The following is no longer correct:
`var streamWriter = streamClient.createWriter("streamName");`

The previous line and the write() calls can be expressed as follows, right?

```
streamClient.createWriter("newStreamName").then(function(writer){
    writer.write("Many messages sent to stream.");
    writer.write("Another message sent to stream.");
    writer.write("Multiple messages sent to stream.");
  })
```

kpovetkin:
It's not outdated. You can do in this way:

var streamWriter = streamClient.createWriter("streamName");
streamWriter.then(function(writer){
    writer.write("Many messages sent to stream.");
    writer.write("Another message sent to stream.");
    writer.write("Multiple messages sent to stream.");
  })

anwar6953:
Right, but the README is currently showing:
`var streamWriter = streamClient.createWriter("streamName");`
and below that its showing
`streamWriter.write("New log event");`
which no longer works.

kpovetkin:
You are right. Updated.

anwar6953:
All else looks good to me.
:+1: 

Fine to merge, unless others have comments.

kpovetkin:
Problem with promises(did not fire handlers in node js console) is solved.

anwar6953:
Looks good to me.

PR #101, opened 2014-11-26
[cdap-524] modified to be possible to use with both python 2 and 3 versions

kpovetkin:
Created according to https://issues.cask.co/browse/CDAP-524

### DIFF
--- a/cdap-stream-clients/python/test/stream_integration_test.py
+++ b/cdap-stream-clients/python/test/stream_integration_test.py
@@ -22,6 +22,7 @@ try:
 except ImportError:
     import unittest as unittest
 import os
+import six
 
 from stream_test_base import StreamTestBase
 
@@ -36,4 +37,8 @@ class TestStreamClient(unittest.TestCase, StreamTestBase):
         self.base_set_up()
 
 if u'__main__' == __name__:
-    unittest.main()
+    if six.PY2:
+        unittest.main()
+
+    if six.PY3:
+        unittest.main(warnings=False)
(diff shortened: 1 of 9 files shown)


anwar6953:
One comment - integration tests don't yet run in python3 (httplib is not on 3.x)

Otherwise looks good to me. Thanks.

kpovetkin:
Updated.

cdap-stream-clients/python/test/stream_integration_test.py
```diff
@@ -36,4 +37,8 @@ def setUp(self):
         self.base_set_up()
 
 if u'__main__' == __name__:
-    unittest.main()
+    if six.PY2:
+        unittest.main()
+
+    if six.PY3:
+        unittest.main(warnings=False)
```
anwar6953: Does it make sense to suppress only specific warnings, so that a develop can still get other meaningful warnings

kpovetkin: It's not possible to suppress specific warnings only. Tests do not produce any usefull warnings at all.

anwar6953: I think its better just to suppress the specific ResourceWarning that you want to suppress:

For instance, if you put this in TestStreamClient.setUp, then you can simply call unittest.main(), without suppressing all other warnings (which there might be useful warnings in the future):
`warnings.simplefilter("ignore", ResourceWarning)`

https://docs.python.org/2/library/warnings.html#temporarily-suppressing-warnings

anwar6953: add a newline to the end of last line of this file

kpovetkin:
Warnings suppression added.

anwar6953:
:+1:

PR #104, opened 2015-03-06
Fix readmes et al

awholegunch:


README.rst
```diff
@@ -1,38 +1,52 @@
...
+::
+
+  mvn clean package -P deb
 
-Build debian for packages (File tailer and DropZone)
-----------------------
-mvn clean package -P deb
 
-For details, please see the README.md file in each submodule.
+For details, please see the README file in each submodule.
```
jwang47: README.rst ?

jwang47:
One minor comment. Otherwise LGTM.

README.rst
awholegunch: If you are under windows and the file extensions are hidden, you won't see it...but...

PR #105, opened 2015-04-16
Port Java stream client and flume sink to v3

anew:
- bumps the version to 1.3.0-SNAPSHOT
- bumps CDAP dependency version to 2.8.0
- changes to Java stream client to support v2 and v3; requires some refactoring
- changes to Flume stream sink to support v3
- updates to integration tests to work with v3

There are no automated tests for this... I tested this manually by running the integration tests for stream client and for the flume sink against a standalone CDAP and against a secure single node cluster.


ghost:
Did not see any update that could alter behavior or flow. +1 LGTM

PR #106, opened 2015-04-17
[CDAP-213] remove file drop zone and file tailer from cdap ingest

anew:
title says it all. 
Oh, this also adds/updates copyright for some files there were missing the correct copyright.


albertshau:
:+1:

awholegunch:
This will break docs. Hang on on merging for a sec...

awholegunch:
Some of these files (README.rst) are currently downloaded by the documentation build scripts, checked against checked in versions, and then built. If they are going away, then we need to change the cdap docs to reflect these changes. That can be done in another PR, but perhaps that could be set up before we merge this one?

awholegunch:
This would change the Developers’ Manual section "Ingesting Data into CDAP". I don't think these items were deprecated before being deleted, were they?

anew:
The docs are in a different repo, so it will have to be a separate PR.
Not sure whether they had been deprecated. We decided to stop supporting them, because we are adding new ETL plugins that make these unnecessary.
Alternatively, we can choose to maintain these tools and migrate them to use v3 APIs. Not sure if we want to spend that effort. 
How much does it take to fix the docs?

awholegunch:
Not much. I can set up the PR...it's more just that by going through deprecation and documentation, that way I learn about it and keep things in sync...not to mention any clients who depend on that. But who should review this section and tell me what all should be revised?:
http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT/en/developers-manual/ingesting-tools/index.html

anew:
For now, you need to remove the sections about file drop zone and file tailer. Also, there is a new configuration called "namespace" that has the obvious semantics. 
I will review it once you are done.

awholegunch:
Have there been any changes in the "Stream Client for Java"? I notice that the example looks "old-school" in its approach to constructing the StreamClient.

anew:
The only change is that you can now specify a namespace, if you use version v3

awholegunch:
README.rst at top level of the repo should be edited to remove the references to the removed projects.

anwar6953:
FileTailer and FileDropZone utilize CDAP Stream Client to interact with streams.
So if CDAP Stream Client has support for v3, shouldn't that mean that FileTailer and FileDropZone also do? (All there needs to be is some way to pass configuration of namespace from FileTailer to the CDAP Stream Client, right?).

Also, while we do have new ETL Plugins, they don't replace the functionality of being able to ingest files (and/or tail+ingest a file/directory) into CDAP, right?

anew:
@anwar6953 the question is also whether we want to support these 9000+ lines of code.

anew:
We decided to keep these, however, we will not port them to v3/ or test them on 3.0, and we will not support them further. Closing this PR and will open a new one to add a note the Readme's that they are no longer supported.

anwar6953:
@anew Ok, that point I can't disagree with.

awholegunch:
Revised PR: https://github.com/caskdata/cdap-ingest/pull/107

PR #107, opened 2015-04-18
[CDAP-2136] deprecate file drop zone and file tailer from cdap ingest

anew:
also adds some missing copyrights.


nitinmotgi:
LGTM :+1:

PR #108, opened 2015-04-21
[CDAP-2129] port python stream client to v3

anew:
- bumps version to 1.3.0 (same as Java stream client)
- adds a namespace to the configuration.
- uses v3 APIs
- removes consumer-id and dequeue as they are not supported in v3
- adapts tests

### DIFF
--- a/cdap-stream-clients/python/cdap_stream_client/config.py
+++ b/cdap-stream-clients/python/cdap_stream_client/config.py
@@ -30,8 +30,8 @@ DEFAULT_VERIFY_SSL_CERT = True
 class Config(object):
 
     def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT,
-                 version=DEFAULT_VERSION, namespace=DEFAULT_NAMESPACE,
-                 ssl=DEFAULT_SSL, verify_ssl_cert=DEFAULT_VERIFY_SSL_CERT):
+                 ssl=DEFAULT_SSL, verify_ssl_cert=DEFAULT_VERIFY_SSL_CERT,
+                 version=DEFAULT_VERSION, namespace=DEFAULT_NAMESPACE):
         self.__host = host
         self.__port = port
         self.__version = version
--- a/cdap-stream-clients/python/cdap_stream_client/config.py
+++ b/cdap-stream-clients/python/cdap_stream_client/config.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-#  Copyright © 2014 Cask Data, Inc.
+#  Copyright © 2014-2015 Cask Data, Inc.
 #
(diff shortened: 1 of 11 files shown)


cdap-stream-clients/python/cdap_stream_client/streamclient.py
```diff
@@ -23,27 +23,22 @@
 
 class StreamClient(ConnectionErrorChecker):
 
-    __GATEWAY_VERSION = u'/v2'
     __REQUEST_PLACEHOLDERS = {
         u'streamid': u'<streamid>'
     }
-    __REQUESTS = {u'streams': __GATEWAY_VERSION + u'/streams'}
+    __REQUESTS = {u'streams': u'/streams'}
```
ghost, line 29: I assume GATEWAY is router in the current architecture? The new code does not need version in the path?

cdap-stream-clients/python/cdap_stream_client/config.py
```diff
@@ -21,16 +21,21 @@
...
 DEFAULT_SSL = False
 DEFAULT_VERIFY_SSL_CERT = True
 
 
 class Config(object):
 
-    def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, ssl=DEFAULT_SSL,
-                 verify_ssl_cert=DEFAULT_VERIFY_SSL_CERT):
+    def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT,
+                 version=DEFAULT_VERSION, namespace=DEFAULT_NAMESPACE,
```
ghost: Probably better to add new arguments at the end of the existing ones to prevent accidental runtime error for existing code calling this constructor?

cdap-stream-clients/python/cdap_stream_client/streamclient.py
anew, line 29: I moved the version to the connection config, along with adding a namespace to the config. This is in parity with the Java stream client.

cdap-stream-clients/python/cdap_stream_client/config.py
anew: sure

anew:
@henry-cask I have addressed your comments. OK to merge?

cdap-stream-clients/python/cdap_stream_client/streamclient.py
anwar6953, line 29: Seems he's passing it via `serviceconnector`, and prepending the version into the path there.

anew, line 29: This is because in v3/ there is a namespace between the version and the stream, so we cannot construct the URL in the same way. The service connector now connects to a v3 namespace, which is part of the connection config.

ghost:
+1

LGTM!

PR #109, opened 2015-04-22
[CDAP-2128] port ruby stream client to v3

anew:
- adds namespace to the config, switches to v3 of the API
- fixes existing tests
- fixes problem with stream-writer (existing code actually was not working, failed on develop, too)
- adds copyright to a bunch of README's


cdap-stream-clients/ruby/lib/stream-client-ruby/version.rb
```diff
@@ -13,5 +13,5 @@
 #  the License.
 
 module CDAPIngest
-  VERSION = "1.1.0"
+  VERSION = "1.3.0"
```
duddun: Use single quotes: VERSION = '1.3.0'

anew: ok.

PR #110, opened 2015-04-22
Cleanup READMEs for CDAP 3.0 release

awholegunch:
Fixes, tweaks, corrections, adjustments. Each file has the same copyright statement at the end. When the file is included elsewhere, the include will stop before it.

### DIFF
--- a/cdap-file-drop-zone/README.rst
+++ b/cdap-file-drop-zone/README.rst
@@ -131,3 +131,23 @@ using 2 pipes::
   pipes.pipe1.sink.verify.ssl.cert=true
   # Path to authentication client properties to use if SSL is being used
   pipes.pipe1.sink.auth_client_properties=/etc/file-drop-zone/conf/auth-client.properties
+  
+
+License and Trademarks
+----------------------
+Copyright © 2014-2015 Cask Data, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+in compliance with the License. You may obtain a copy of the License at
+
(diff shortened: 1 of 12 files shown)


cdap-file-drop-zone/README.rst
```diff
@@ -1,3 +1,8 @@
+.. meta::
+    :author: Cask Data, Inc.
+    :copyright: Copyright © 2015 Cask Data, Inc.
+    :license: See LICENSE file in this repository
+
```
anew, line 5: Copyright is 2014-2015, right? Because these were written in 2014...

awholegunch:
Fixed copyright.

cdap-stream-clients/java/README.rst
```diff
@@ -60,7 +65,7 @@ Example::
 
 Creating a Stream
 -----------------
-Create a new Stream with the *stream id* "streamName"::
+Create a new Stream with the *stream-id* "streamName"::
```
anew: Create a new Stream named `streamName` ?

anew: Hmm not sure what the formatting should be... what is the convention elsewhere? What is italic used for in other places?

anew:
LGTM

cdap-stream-clients/java/README.rst
awholegunch: it's probably backward. Names are in italics. Parameters are in between angle brackets, and in literal. Let me fix these...

PR #115, opened 2016-10-12
Add codeclimate configuration

wolf31o2:
This is the default configuration other than excluding testing our tests. Tests commonly have duplicated code and other such things that we want to fix in the main codebase, but aren't very concerned about in tests.

screen shot 2016-10-12 at 5 04 25 pm


wolf31o2:
Closing this

